PR #18, opened 2014-02-12
Don't escape method filter regexp

csfrancis:
The regular expression value passed to `--method` is being incorrectly escaped. This leads to the inability to perform searches for things like `foo.*bar`, because the dot character is escaped and the resulting regexp becomes `foo\.*bar`.


pushrax:
@tmm1 ping

PR #39, opened 2014-11-04
Add raw option for StackProf::Middleware

SpringMT:
Add raw option for StackProf::Middleware.

When this option is enabled, the report can convert flamegraph and so on.

Usage is below.

```
use StackProf::Middleware, enabled: true,
                           mode: :cpu,
                           interval: 1000,
                           save_every: 5,
                           raw: true
```

### DIFF
--- a/lib/stackprof/middleware.rb
+++ b/lib/stackprof/middleware.rb
@@ -9,6 +9,7 @@ module StackProf
 
       Middleware.mode     = options[:mode] || :cpu
       Middleware.interval = options[:interval] || 1000
+      Middleware.raw      = options[:raw] || false
       Middleware.enabled  = options[:enabled]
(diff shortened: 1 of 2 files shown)


Sirupsen:
@SpringMT think you need to pass in `raw: Middleware.raw` to `Start`.

TheDude05:
+1 to @Sirupsen comments as well as the PR in general. I was looking for this functionality today.

tmm1:
Oops merged without the fix to pass the value through. Could someone open another PR with that one liner?

Sirupsen:
@tmm1 https://github.com/tmm1/stackprof/pull/48

SpringMT:
@tmm1 @Sirupsen  thx!.

PR #41, opened 2014-11-19
Add option to drop less important nodes for big graphviz output

### LINKED ISSUE #40: Drop edges/nodes below a given count
pprof provides a useful graph feature where you can selectively drop edges and nodes below a certain sample count. Trying to profile a huge app can result in tens of thousands of connections, which easily kills `dot` when rendering a graph.

pushrax:
Fixes #40.

Just a basic implementation of node dropping based on total_samples. I chose this metric in order to preserve the call chain in expensive paths, rather than dropping intermediate nodes.

No tests included since there don't seem to be any existing ones for the output functions.

@camilo @csfrancis fyi

### DIFF
--- a/bin/stackprof
+++ b/bin/stackprof
@@ -2,23 +2,20 @@
 require 'optparse'
 require 'stackprof'
 
-options = {
-  :format => :text,
-  :sort => false,
-  :limit => 30
-}
+options = {}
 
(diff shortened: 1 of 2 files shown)


pushrax:
@tmm1 ping

PR #44, opened 2015-01-06
Pass optional io to Report#print_dump

splattael:
After this change you can dump your profile results to any IO via:

```ruby
report = StackProf::Report.new(StackProf.results)
File.open("profile.dump", "wb") do |io|
  report.print_dump(io)
end
```


splattael:
@tmm1 Yay, thanks for merging! :)

PR #46, opened 2015-03-25
Fix filtering for Graphviz report

splattael:
After this commit passing a `filter` to `print_graphviz` works again.


splattael:
Sorry, had to rename the branch :/

PR #56, opened 2016-01-04
Add `.travis.yml`

yui-knk:


yui-knk:
See also https://travis-ci.org/yui-knk/stackprof

PR #59, opened 2016-01-25
Avoid "stack level too deep" error

yuroyoro:
When too many frames are included in the data,  Stackprof raises `SystemStackError(stack level too deep)` at constructing the Hash object from `@data[:frames]`.

```
/home/ozaki/dev/asp_weborder/vendor/bundle/ruby/2.2.0/gems/stackprof-0.2.7/lib/stackprof/report.rb:12:in `frames': stack level too deep (SystemStackError)
        from /home/ozaki/dev/asp_weborder/vendor/bundle/ruby/2.2.0/gems/stackprof-0.2.7/lib/stackprof/report.rb:199:in `print_text'
        from /home/ozaki/dev/asp_weborder/vendor/bundle/ruby/2.2.0/gems/stackprof-0.2.7/bin/stackprof:48:in `<top (required)>'
        from /home/ozaki/dev/asp_weborder/vendor/bundle/ruby/2.2.0/bin/stackprof:23:in `load'
        from /home/ozaki/dev/asp_weborder/vendor/bundle/ruby/2.2.0/bin/stackprof:23:in `<main>'
```

For avoiding this error, I rewrote `frames` method by `Hash#inject` instead of using `Hash[ *@data[:frames] ...]`  

F.Y.I 
https://bugs.ruby-lang.org/issues/5719

### DIFF
--- a/lib/stackprof/report.rb
+++ b/lib/stackprof/report.rb
@@ -9,7 +9,7 @@ module StackProf
     attr_reader :data
 
     def frames(sort_by_total=false)
-      Hash[ *@data[:frames].sort_by{ |iseq, stats| -stats[sort_by_total ? :total_samples : :samples] }.flatten(1) ]
+      *@data[:frames].sort_by{ |iseq, stats| -stats[sort_by_total ? :total_samples : :samples] }.inject({}){|h, (k, v)| h[k] = v; h}
     end
 
     def normalized_frames


tmm1:
Thank you!

lib/stackprof/report.rb
```diff
@@ -9,7 +9,7 @@ def initialize(data)
     attr_reader :data
 
     def frames(sort_by_total=false)
-      Hash[ *@data[:frames].sort_by{ |iseq, stats| -stats[sort_by_total ? :total_samples : :samples] }.flatten(1) ]
+      *@data[:frames].sort_by{ |iseq, stats| -stats[sort_by_total ? :total_samples : :samples] }.inject({}){|h, (k, v)| h[k] = v; h}
```
seanlinsley: Doesn't this cause a syntax error?
```ruby
2.2.3 :001 > def a
2.2.3 :002?>   *[1, 2]
2.2.3 :003?> end
SyntaxError: (irb):1: syntax error, unexpected '\n', expecting :: or '[' or '.'
	from ~/.rvm/rubies/ruby-2.2.3/bin/irb:15:in `<main>'
```

It appears that stackprof isn't set up to run its tests on a CI service. @tmm1 is that something you'd want? I'd open a PR myself, except I'm pretty sure only people with commit access can configure Travis and the like.

tmm1: I will enable travis if you add the config in a PR.

seanlinsley: Why was `flatten(1)` removed?

tmm1: Doesn't look like there's good test coverage for the reporter anyway. I should have reviewed this more carefully, will revert for now.

seanlinsley: In Ruby 2.2.3 this code works (while it takes a few minutes and takes 7 GB of memory):
```ruby
Hash[60_000_000.times.map{ |n| [n, n] }].size
 => 60000000
```
Perhaps it's the `*` that makes the difference? I'm not familiar with the data structure, so I'm not sure why the destructuring is needed.

seanlinsley: BTW, `to_h` seems to have essentially the same performance as `Hash[]`

yuroyoro: Sorry, I made a mistake. `*` isn't needed.
I fixed syntax error. 
May I send PR again?

tmm1: Yes please

PR #69, opened 2016-07-06
Update readme with flamegraph example

schneems:
Took me forever to actually figure out how the commands to generate an view a flamegraph were supposed to be used. I added an example.

Also wanted to give an example on how to set interval option in the same section where we're talking about default intervals.

Added a `#` before `...` so it's technically valid ruby code.

Uppercase section header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ in ruby:
 
 ``` ruby
 StackProf.run(mode: :cpu, out: 'tmp/stackprof-cpu-myapp.dump') do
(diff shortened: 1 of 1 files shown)


schneems:
Btw, thanks for Stackprof. I've made some pretty good gains on Sprockets 3 and 4 with it. First time i've ever used it this much.

seanlinsley:
Thank you for improving the documentation ❤️

Issue #73, opened 2016-10-19
Confused about flamegraph generation


waiting-for-dev:
Hi,

I've been trying to generate a flamegraph with stackprof.

First, I've been following what it seems the recommended way, because it is documented in the README:

    stackprof --flamegraph tmp/stackprof-cpu-myapp.dump > tmp/flamegraph
    `stackprof --flamegraph-viewer=tmp/flamegraph`

However, this always resulted in a web page that got unresponsive due to what it seems excessive javascript processing. I had to increase `interval` option to unacceptable values to generate a very small flamegraph file which could be processed by the browser (and even that way the graphic was very small, unreadable).

Then, digging a bit I found out there is another way to generate the flamegraph, which, if I'm not wrong, uses a https://github.com/brendangregg/FlameGraph tool (which is included in stackprof in vendor directory):

    stackprof --stackcollapse tmp/stackprof-cpu-myapp.dump > tmp/flamegraph
    stackprof-flamegraph.pl flamegraph > tmp/flamegraph.svg
    firefox tmp/flamegraph.svg

This solution, even if does not render any legend and it is just the svg without any js, works really well.

Do you think that, at least for now, this should be the recommended way? I don't mind submitting a request with an update to the README file.

Thanks for your work in this awesome gem :)

tmm1:
The JavaScript flamegraph is preferred, since its interactive and has more details. You should be able to zoom in and out.

What browser are you using?

@aroben Maybe your perf patches to flamegraph.js would help here?

waiting-for-dev:
Thanks for your response. I tried with both firefox and chrome in debian.

aroben:
I'm actually working on a branch right now that replaces the current SVG/D3 Flamegraph viewer with one based on `<canvas>`. It's far more responsive than the current viewer for large profiles. I'm hoping to get a PR opened up in the next day or two with the new implementation.

waiting-for-dev:
Very happy to hear that :D

PR #74, opened 2016-10-19
Rewrite flamegraph viewer using canvas

aroben:
Large profiles were generating so many SVG nodes that the browser was
bogging down. Using canvas keeps the DOM very simple, and repainting it
seems to be pretty fast. It also lets us drop the D3 and jQuery
dependencies for a smaller page load.

The viewer has been flipped vertically: the small overview graph is on
the top, and stacks grow top down instead of bottom up. This matches
Chrome's devtools timeline and many other timeline/flamegraph tools I've
seen.

Controls are:
- drag: pan/scroll
- scrollwheel: pan/scroll
- Shift-scrollwheel: zoom

Click-and-drag to select a region in the small overview graph isn't
currently supported, nor is dragging the selected region to pan/scroll.
That would be nice to add back eventually, though I've found that
zoom/scroll performance is good enough that I haven't really missed
those features.

This should be compatible with IE 10 and newer, though I've only tested
in Safari, Firefox, and Chrome.

/cc #73 @waiting-for-dev @tmm1 @kdaigle

### DIFF
--- a/lib/stackprof/flamegraph/flamegraph.js
+++ b/lib/stackprof/flamegraph/flamegraph.js
@@ -439,6 +439,10 @@ function OverviewFlamechart(container, viewportOverlay, data, dataRange, info) {
   self.canvas.addEventListener('mousedown', self.onMouseDown.bind(self))
   self.canvas.addEventListener('mouseup', self.onMouseUp.bind(self))
   self.canvas.addEventListener('mousemove', self.onMouseMove.bind(self))
+
+  self.viewportOverlay.addEventListener('mousedown', self.onOverlayMouseDown.bind(self))
+  self.viewportOverlay.addEventListener('mouseup', self.onOverlayMouseUp.bind(self))
+  self.viewportOverlay.addEventListener('mousemove', self.onOverlayMouseMove.bind(self))
 }
 
 OverviewFlamechart.prototype = Object.create(Flamechart.prototype)
@@ -513,6 +517,62 @@ OverviewFlamechart.prototype.handleDragGesture = function(e) {
   rect.width = Math.max(self.viewport.width / 1000, (maxX - minX) / self.width * self.viewport.width)
 
   rect.y = Math.max(self.viewport.y, Math.min(self.viewport.height - self.viewport.y, e.offsetY / self.height * self.viewport.height + self.viewport.y - rect.height / 2))
+
+  self.setViewportOverlayRect(rect)
+  self.dispatch('overlaychanged', { current: self.viewportOverlayRect })
+}
+
+OverviewFlamechart.prototype.onOverlayMouseDown = function(e) {
+  var self = this
+
+  self.overlayDragging = true
+  self.overlayDragInfo = {
+    mouse: { x: e.clientX, y: e.clientY },
+    rect: Object.assign({}, self.viewportOverlayRect),
+  }
+  self.viewportOverlay.classList.add('moving')
+
+  self.handleOverlayDragGesture(e)
+
+  e.preventDefault()
+}
+
+OverviewFlamechart.prototype.onOverlayMouseUp = function(e) {
+  var self = this
+
+  if (!self.overlayDragging)
+    return
+
+  self.overlayDragging = false
+  self.viewportOverlay.classList.remove('moving')
+
+  self.handleOverlayDragGesture(e)
+
+  e.preventDefault()
+}
+
+OverviewFlamechart.prototype.onOverlayMouseMove = function(e) {
+  var self = this
+
+  if (!self.overlayDragging)
+    return
+
+  self.handleOverlayDragGesture(e)
+
+  e.preventDefault()
+}
+
+OverviewFlamechart.prototype.handleOverlayDragGesture = function(e) {
+  var self = this
+
+  var deltaX = (e.clientX - self.overlayDragInfo.mouse.x) / self.width * self.viewport.width
+  var deltaY = (e.clientY - self.overlayDragInfo.mouse.y) / self.height * self.viewport.height
+
+  var rect = Object.assign({}, self.overlayDragInfo.rect)
+  rect.x += deltaX
+  rect.y += deltaY
+  rect.x = Math.max(self.viewport.x, Math.min(self.viewport.x + self.viewport.width - rect.width, rect.x))
+  rect.y = Math.max(self.viewport.y, Math.min(self.viewport.y + self.viewport.height - rect.height, rect.y))
 
   self.setViewportOverlayRect(rect)
   self.dispatch('overlaychanged', { current: self.viewportOverlayRect })
--- a/lib/stackprof/flamegraph/flamegraph.js
+++ b/lib/stackprof/flamegraph/flamegraph.js
@@ -1,357 +1,787 @@
-var guessGem = function(frame) {
-  var split = frame.split('/gems/');
-  if(split.length == 1) {
-    split = frame.split('/app/');
-    if(split.length == 1) {
-      split = frame.split('/lib/');
-    } else {
-      return split[split.length-1].split('/')[0]
+if (typeof Element.prototype.matches !== 'function') {
+  Element.prototype.matches = Element.prototype.msMatchesSelector || Element.prototype.mozMatchesSelector || Element.prototype.webkitMatchesSelector || function matches(selector) {
+    var element = this
+    var elements = (element.document || element.ownerDocument).querySelectorAll(selector)
+    var index = 0
+
+    while (elements[index] && elements[index] !== element) {
+      ++index
     }
 
-    split = split[Math.max(split.length-2,0)].split('/');
-    return split[split.length-1].split(':')[0];
+    return Boolean(elements[index])
+  }
+}
+
+if (typeof Element.prototype.closest !== 'function') {
+  Element.prototype.closest = function closest(selector) {
+    var element = this
+
+    while (element && element.nodeType === 1) {
+      if (element.matches(selector)) {
+        return element
+      }
+
+      element = element.parentNode
+    }
+
+    return null
+  }
+}
+
+if (typeof Object.assign !== 'function') {
+  (function() {
+    Object.assign = function(target) {
+      'use strict'
+      // We must check against these specific cases.
+      if (target === undefined || target === null) {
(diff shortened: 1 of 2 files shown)


aroben:
I've set up an example so you can try it out.

Current behavior from `master`: http://roben.org/stackprof-flamegraph/b4e5d825e057cb25bbc72b538219643d3d5e7316/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt

Behavior from this branch: http://roben.org/stackprof-flamegraph/6dcff3a43ee18b6a7162afce24fcbd171e354ed5/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt

tmm1:
Looks great! Definitely many times more performant.

cc @samsaffron

SamSaffron:
It is very cool, some bits of feedback... 

- I would be nice to be able to select range from the top part like chrome does with flamechart
- frame info is missing, which is very useful (click on frame) 
- would love if this is instead added to https://github.com/SamSaffron/flamegraph and then stackprof consumes that instead (cause rack-mini-profiler uses flamegraph gem and it is a standalone thing)

waiting-for-dev:
Hey, great job! Just two considerations:

- For a lot of data it gets pretty quirk. Following screenshot if with 88M:

screenshot-flamegraph

- Scrolling to the bottom seems to be quite slow

Many, many thanks!

aroben:
@SamSaffron @waiting-for-dev Thanks for trying it out.

> I would be nice to be able to select range from the top part like chrome does with flamechart

This was mentioned as a limitation in the PR body, and I agree it would be good to add. Since you mentioned it I'll try to get it implemented in this initial version.

> frame info is missing, which is very useful (click on frame)

Is this different from the info that is shown when you hover over a frame?

screen shot 2016-10-20 at 9 13 30 am

> would love if this is instead added to https://github.com/SamSaffron/flamegraph and then stackprof consumes that instead (cause rack-mini-profiler uses flamegraph gem and it is a standalone thing)

@tmm1 what do you think about this?

> For a lot of data it gets pretty quirk. Following screenshot if with 88M:

@waiting-for-dev I don't see anything obviously wrong with that screenshot. To me it just looks like you need to zoom in (Shift-mousewheel) so you can see what's going on in those lower regions. Have you tried that?

> Scrolling to the bottom seems to be quite slow

If you can provide me with your stackprof profile I can try to figure out how to speed it up.

aroben:
I've noticed that scrolling performance is much better in Safari than in Chrome. I guess Safari's canvas is more optimized for this case. I'll see if I can figure out how to make it faster in Chrome. (It's still far faster than the SVG implementation.)

aroben:
Looks like `CanvasRenderingContext2D.prototype.measureText` is much slower in Chrome than in Safari, which accounts for almost all of the performance difference. I'll see if I can minimize calls to that method.

aroben:
I just pushed up some optimizations. Rendering time for a 3MB profile increased from 9.9fps to 50fps in Chrome, and from 41.7fps to 62.5fps in Safari. Check it out at http://roben.org/stackprof-flamegraph/3381fd86cf0e01918f4f5786481e724e4682fd58/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt

aroben:
> I would be nice to be able to select range from the top part like chrome does with flamechart

I've pushed up some support for this.

aroben:
With mouse support for the top part: http://roben.org/stackprof-flamegraph/2cd647b4c422971c6293a7f8af4a4cf3ff8857a6/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt

aroben:
I'd like to implement capturing of mouse events while dragging. Right now you can quickly move the mouse off of the overlay while you're dragging and it will stop responding to you.

SamSaffron:
Getting lots of artifacts on the latest example:

image

Regarding frame info was talking about this:

image

Long term I want us to ship an optional middleware that allows you to view a syntax highlighted source file, that way when you see a frame is playing up you can simply click and view source. 

This also highlight the problem of having a fork of flamegraph, everyone end up losers this way. 

@tmm1 I think it is correct to keep it in flamegraph gem, that way @headius can eventually make this thing work in JRuby as well :) 

Happy to give you and @aroben ownership on flamegraph gem, in fact I did so now.

aroben:
> Getting lots of artifacts on the latest example:

I seem to have really screwed things up on Retina displays. I was working on a non-Retina display today. This shouldn't be too hard to fix. I'll take a look tomorrow.

> Regarding frame info was talking about this:

Ah, interesting. That does seem nice. The existing implementation in this repo doesn't have that view, so I think this PR could probably land without it. But if we move this into the flamegraph gem we'd of course want to keep that working.

> Long term I want us to ship an optional middleware that allows you to view a syntax highlighted source file, that way when you see a frame is playing up you can simply click and view source.

Yes, having a quick way to get from the flamegraph to the source code is something I've been wanting as well.

waiting-for-dev:
Thanks for your work on this @aroben .

> @waiting-for-dev I don't see anything obviously wrong with that screenshot. To me it just looks like you need to zoom in (Shift-mousewheel) so you can see what's going on in those lower regions. Have you tried that?

You are right. I can zoom in. I had tried with Firefox and, there, now I see the zoom works in the opposite way I feel natural: I need to move the mousewheel backwards. In Chrome it works "well"; forwards. I don't know if it is just some browser default behavior or something.

> If you can provide me with your stackprof profile I can try to figure out how to speed it up.

Now I see that, in Chrome, scrolling and zooming have an acceptable speed. However, in Firefox is quite slow, making it not very handy to work with large files. In a minute, I'm going to send to your private email account (the one in your Github profile) a link to download a real example.

aroben:
> You are right. I can zoom in. I had tried with Firefox and, there, now I see the zoom works in the opposite way I feel natural: I need to move the mousewheel backwards. In Chrome it works "well"; forwards. I don't know if it is just some browser default behavior or something.

What OS and browser versions are you using? If you're on OS X, do you have "Scroll direction: natural" enabled in your mouse settings?

> Now I see that, in Chrome, scrolling and zooming have an acceptable speed. However, in Firefox is quite slow, making it not very handy to work with large files. In a minute, I'm going to send to your private email account (the one in your Github profile) a link to download a real example.

I'll take a look. Thanks!

waiting-for-dev:
> What OS and browser versions are you using? If you're on OS X, do you have "Scroll direction: natural" enabled in your mouse settings?

I'm on Firefox 48.0.2 in debian 8 (in fact a VM on QubesOS 3.2, but this should be irrelevant).

aroben:
> I'm on Firefox 48.0.2 in debian 8 (in fact a VM on QubesOS 3.2, but this should be irrelevant).

Were you using Chrome on a Mac, or also on Linux? If Chrome and Firefox were both on Linux I'd be surprised if there zoom behavior were different.

http://roben.org/stackprof-flamegraph/df8afd35e9c7f85a92532a7323682a05d55c4a35/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt should have the zoom direction fixed on Windows/Linux. Let me know how it works for you.

aroben:
> > Now I see that, in Chrome, scrolling and zooming have an acceptable speed. However, in Firefox is quite slow, making it not very handy to work with large files. In a minute, I'm going to send to your private email account (the one in your Github profile) a link to download a real example.
> 
> I'll take a look. Thanks!

Firefox and Chrome perform about the same for me with that profile. I'm on macOS 10.12 and don't have a Linux machine handy for testing. You could try capturing a JavaScript profile in Firefox's devtools and see where the time is being spent.

waiting-for-dev:
Hey! I'll be able to look at this on Monday. This weekend I'm far from the computer :) Thanks!

waiting-for-dev:
> Were you using Chrome on a Mac, or also on Linux? If Chrome and Firefox were both on Linux I'd be surprised if there zoom behavior were different.

Yes, it was in the same Linux machine. It also seemed strange to me...

> http://roben.org/stackprof-flamegraph/df8afd35e9c7f85a92532a7323682a05d55c4a35/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt should have the zoom direction fixed on Windows/Linux. Let me know how it works for you.

Great! Now in both Firefox and Chrome I have to move the mousewheel forwards.

> Firefox and Chrome perform about the same for me with that profile. I'm on macOS 10.12 and don't have a Linux machine handy for testing. You could try capturing a JavaScript profile in Firefox's devtools and see where the time is being spent.

It is not that it performs bad in Firefox. Simply it has an unsuitable "scale". I mean, for the same mousewheel movement I zoom in much more in Chrome than in Firefox. Anyway, here you have a javascript profile just zooming in in Firefox:

https://gist.github.com/waiting-for-dev/4b7b241cc330ee33c68b8af670192a88

Thanks

waiting-for-dev:
Hey, I'll be able to look at this on Monday. I'm far from the computer this
weekend :) Thanks!

aroben:
@waiting-for-dev I think I've fixed the Firefox issue. Can you try http://roben.org/stackprof-flamegraph/152c81f86051e619df1d57f1cfb84924c62938b3/viewer.html?data=/stackprof-flamegraph/rubocop-spec-flamegraph.txt

waiting-for-dev:
@aroben awesome! Now works perfectly well in Firefox :)

parkr:
Hey! I was just trying to view a flamegraph from stackprof and got quite stuck. My browser locked up completely. I was looking for one of two things:

1. A more performant alternative to the viewer (like this PR!)
2. Output to a `flamegraph.pl`-compatible input format.

@aroben It sounds like you have addressed all comments above, and this would greatly benefit folks using this gem. Are there any other tests you'd like to run? Perhaps this could be merged into a v0.3.0 of stackprof?

aroben:
I don't have any objections to merging this (and as far as I know it's been working well internally at GitHub) but I don't have permission to merge it myself.

aroben:
@tmm1 are you able to make a new release?

parkr:
Thanks @aroben and @tmm1! There are a lot of really excellent projects under your account, @tmm1! Thank you for sharing them 😄

brianmario:
@aroben @parkr just FYI, I think he's out of town for at least another week I think.

PR #77, opened 2017-01-19
fix total sample accounting in case of recursive calls

doudou:
When there is a recursive call in the backtrace, the sample would
be accounted twice in total_count.


doudou:
Ping

tmm1:
Sorry I missed the PR. This has been a long-standing bug, so happy to see a fix.

Is it possible to add a test?

doudou:
> Is it possible to add a test?

I'll try.

PR #87, opened 2017-11-16
Add fake stack frames representing samples taken during garbage collection

### LINKED ISSUE #86: Include gc samples in raw output
Hi there,

I'm working on tooling to convert stackprof output into something consumable by chrome developer tools to generate nice interactive flamegraphs.

image

When I was checking the results, I noticed a significant discrepancy between the total sampling profile time and the total wall time elapsed (I am use the `wall` method of sampling). It seems like this is attributable to 2 things: missed samples and gc samples.

It would be nice to include `gc` samples in the `raw` output so they can have corresponding sections of the flamegraph generated for them. I think this is doable by generating a fake stack frame that corresponds to garbage collection. This would help the generated flame charts align more closely with wall time.

If this is something you'd be interested in, I'm happy to try to submit a PR for this.

On a related note, it would be handy to have the walltime of each sample in order to accommodate for missed samples. At the moment, as far as I can tell, it's not possible to identify where in the raw output a missed sample occurred.

P.S. I know there's already flamegraph support via the `flamegraph` gem, but I find the chrome developer tools version much friendlier, and it includes interactive bottoms-up and call-tree methods of inspecting the profiling results.

jlfwong:
When we attempt to take a sample during a garbage collection, instead of just incrementing a counter, we can include a sample in the `raw` output corresponding to a fake stake frame indicating that we're in the middle of a garbage collection. In this case, we include a fake name of `(garbage collection)` and an empty file name.

Fixes #86

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -43,6 +43,7 @@ static struct {
     size_t during_gc;
     st_table *frames;
 
+    VALUE fake_gc_frame;
     VALUE frames_buffer[BUF_SIZE];
     int lines_buffer[BUF_SIZE];
 } _stackprof;
@@ -187,16 +188,24 @@ frame_i(st_data_t key, st_data_t val, st_data_t arg)
 
     rb_hash_aset(results, rb_obj_id(frame), details);
 
-    name = rb_profile_frame_full_label(frame);
+    if (frame == _stackprof.fake_gc_frame) {
+	name = rb_str_new_literal("(garbage collection)");
+	file = rb_str_new_literal("");
+	line = INT2FIX(0);
+    } else {
+	name = rb_profile_frame_full_label(frame);
+
+	file = rb_profile_frame_absolute_path(frame);
+	if (NIL_P(file))
+	    file = rb_profile_frame_path(frame);
+	line = rb_profile_frame_first_lineno(frame);
+    }
+
     rb_hash_aset(details, sym_name, name);
-
-    file = rb_profile_frame_absolute_path(frame);
-    if (NIL_P(file))
-	file = rb_profile_frame_path(frame);
     rb_hash_aset(details, sym_file, file);
-
-    if ((line = rb_profile_frame_first_lineno(frame)) != INT2FIX(0))
+    if (line != INT2FIX(0)) {
 	rb_hash_aset(details, sym_line, line);
+    }
 
     rb_hash_aset(details, sym_total_samples, SIZET2NUM(frame_data->total_samples));
     rb_hash_aset(details, sym_samples, SIZET2NUM(frame_data->caller_samples));
@@ -340,13 +349,12 @@ st_numtable_increment(st_table *table, st_data_t key, size_t increment)
 }
 
 void
-stackprof_record_sample()
+stackprof_record_sample_for_stack(int num)
 {
-    int num, i, n;
+    int i, n;
     VALUE prev_frame = Qnil;
 
     _stackprof.overall_samples++;
-    num = rb_profile_frames(0, sizeof(_stackprof.frames_buffer) / sizeof(VALUE), _stackprof.frames_buffer, _stackprof.lines_buffer);
 
     if (_stackprof.raw) {
 	int found = 0;
@@ -411,6 +419,14 @@ stackprof_record_sample()
     }
 }
 
+void
+stackprof_record_sample()
+{
+    int num = rb_profile_frames(0, sizeof(_stackprof.frames_buffer) / sizeof(VALUE), _stackprof.frames_buffer, _stackprof.lines_buffer);
+    stackprof_record_sample_for_stack(num);
+}
+
+
 static void
 stackprof_job_handler(void *data)
 {
@@ -427,10 +443,14 @@ static void
 stackprof_signal_handler(int sig, siginfo_t *sinfo, void *ucontext)
 {
     _stackprof.overall_signals++;
-    if (rb_during_gc())
-	_stackprof.during_gc++, _stackprof.overall_samples++;
-    else
+    if (rb_during_gc()) {
+	_stackprof.during_gc++;
+	_stackprof.frames_buffer[0] = _stackprof.fake_gc_frame;
+	_stackprof.lines_buffer[0] = -1;
+	stackprof_record_sample_for_stack(1);
+    } else {
 	rb_postponed_job_register_one(0, stackprof_job_handler, 0);
+    }
 }
 
 static void
@@ -530,6 +550,7 @@ Init_stackprof(void)
 #undef S
 
     gc_hook = Data_Wrap_Struct(rb_cObject, stackprof_gc_mark, NULL, &_stackprof);
+    _stackprof.fake_gc_frame = rb_str_new_literal("fake_gc_frame");
     rb_global_variable(&gc_hook);
 
     rb_mStackProf = rb_define_module("StackProf");
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -552,6 +552,7 @@ Init_stackprof(void)
 #undef S
 
     gc_hook = Data_Wrap_Struct(rb_cObject, stackprof_gc_mark, NULL, &_stackprof);
+    rb_global_variable(&gc_hook);
 
     _stackprof.fake_gc_frame = rb_str_new_literal("fake_gc_frame");
     _stackprof.empty_string = rb_str_new_literal("");
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -366,7 +366,7 @@ stackprof_record_sample_for_stack(int num)
 	    _stackprof.raw_samples = malloc(sizeof(VALUE) * _stackprof.raw_samples_capa);
 	}
 
-	if (_stackprof.raw_samples_capa <= _stackprof.raw_samples_len + num) {
+	while (_stackprof.raw_samples_capa <= _stackprof.raw_samples_len + (num + 2)) {
 	    _stackprof.raw_samples_capa *= 2;
 	    _stackprof.raw_samples = realloc(_stackprof.raw_samples, sizeof(VALUE) * _stackprof.raw_samples_capa);
 	}
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -44,6 +44,8 @@ static struct {
     st_table *frames;
 
     VALUE fake_gc_frame;
+    VALUE fake_gc_frame_name;
+    VALUE empty_string;
     VALUE frames_buffer[BUF_SIZE];
     int lines_buffer[BUF_SIZE];
 } _stackprof;
@@ -189,8 +191,8 @@ frame_i(st_data_t key, st_data_t val, st_data_t arg)
     rb_hash_aset(results, rb_obj_id(frame), details);
 
     if (frame == _stackprof.fake_gc_frame) {
-	name = rb_str_new_literal("(garbage collection)");
-	file = rb_str_new_literal("");
+	name = _stackprof.fake_gc_frame_name;
+	file = _stackprof.empty_string;
 	line = INT2FIX(0);
     } else {
 	name = rb_profile_frame_full_label(frame);
@@ -550,8 +552,13 @@ Init_stackprof(void)
 #undef S
 
     gc_hook = Data_Wrap_Struct(rb_cObject, stackprof_gc_mark, NULL, &_stackprof);
+
     _stackprof.fake_gc_frame = rb_str_new_literal("fake_gc_frame");
-    rb_global_variable(&gc_hook);
+    _stackprof.empty_string = rb_str_new_literal("");
+    _stackprof.fake_gc_frame_name = rb_str_new_literal("(garbage collection)");
+    rb_global_variable(&_stackprof.fake_gc_frame);
+    rb_global_variable(&_stackprof.fake_gc_frame_name);
+    rb_global_variable(&_stackprof.empty_string);
 
     rb_mStackProf = rb_define_module("StackProf");
     rb_define_singleton_method(rb_mStackProf, "running?", stackprof_running_p, 0);
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -579,8 +579,8 @@ Init_stackprof(void)
     rb_global_variable(&gc_hook);
 
     _stackprof.fake_gc_frame = INT2FIX(0x9C);
-    _stackprof.empty_string = rb_str_new_literal("");
-    _stackprof.fake_gc_frame_name = rb_str_new_literal("(garbage collection)");
+    _stackprof.empty_string = rb_str_new_cstr("");
+    _stackprof.fake_gc_frame_name = rb_str_new_cstr("(garbage collection)");
     rb_global_variable(&_stackprof.fake_gc_frame_name);
     rb_global_variable(&_stackprof.empty_string);
 
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -41,6 +41,7 @@ static struct {
     size_t overall_signals;
     size_t overall_samples;
     size_t during_gc;
+    size_t unrecorded_gc_samples;
     st_table *frames;
 
     VALUE fake_gc_frame;
@@ -429,26 +430,40 @@ stackprof_record_sample()
 }
 
 void
-stackprof_record_gc_sample()
+stackprof_record_gc_samples()
 {
+    int i;
+
     _stackprof.frames_buffer[0] = _stackprof.fake_gc_frame;
     _stackprof.lines_buffer[0] = 0;
-    stackprof_record_sample_for_stack(1);
+    for (i = 0; i < _stackprof.unrecorded_gc_samples; i++) {
+	stackprof_record_sample_for_stack(1);
+    }
+    _stackprof.during_gc += _stackprof.unrecorded_gc_samples;
+    _stackprof.unrecorded_gc_samples = 0;
 }
 
 static void
-stackprof_job_handler(void *is_gc)
+stackprof_gc_job_handler(void *data)
 {
     static int in_signal_handler = 0;
     if (in_signal_handler) return;
     if (!_stackprof.running) return;
 
     in_signal_handler++;
-    if ((int)is_gc) {
-	stackprof_record_gc_sample();
-    } else {
-	stackprof_record_sample();
-    }
+    stackprof_record_gc_samples();
+    in_signal_handler--;
+}
+
+static void
+stackprof_job_handler(void *data)
+{
+    static int in_signal_handler = 0;
+    if (in_signal_handler) return;
+    if (!_stackprof.running) return;
+
+    in_signal_handler++;
+    stackprof_record_sample();
     in_signal_handler--;
 }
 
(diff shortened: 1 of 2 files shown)


jlfwong:
I'm not totally sure how to test this properly -- the test I updated passes, but there are tests that fail before and after my changes for me.

I'm running `bundle exec rake test`. My ruby version is:

```
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin15]
```

Here are the tests that fail for me:

```
  1) Failure:
StackProfTest#test_object_allocation [/Users/jlfwong/code/stackprof/test/test_stackprof.rb:44]:
--- expected
+++ actual
@@ -1 +1 @@
-"block in StackProfTest#test_object_allocation"
+"StackProfTest#test_object_allocation"



  2) Failure:
StackProfTest#test_cputime [/Users/jlfwong/code/stackprof/test/test_stackprof.rb:68]:
Expected: "block in StackProfTest#math"
  Actual: "StackProfTest#math"


  3) Failure:
StackProfTest#test_raw [/Users/jlfwong/code/stackprof/test/test_stackprof.rb:108]:
--- expected
+++ actual
@@ -1 +1 @@
-"block (2 levels) in StackProfTest#test_raw"
+"StackProfTest#test_raw"



  4) Failure:
StackProfTest#test_custom [/Users/jlfwong/code/stackprof/test/test_stackprof.rb:93]:
--- expected
+++ actual
@@ -1 +1 @@
-"block (2 levels) in StackProfTest#test_custom"
+"StackProfTest#test_custom"

```

There are also various ways in which the code seems to crash with my changes:

```
..........FF...F../Users/jlfwong/code/stackprof/test/test_stackprof.rb:125: [BUG] Illegal instruction at 0x007fffd288ab44
```

and

```
.........F.FF.FF/Users/jlfwong/code/stackprof/test/test_stackprof.rb:125: [BUG] rb_gc_mark(): 0x007fed6f9172d8 is T_NONE
```

ext/stackprof/stackprof.c
```diff
@@ -530,6 +550,7 @@ Init_stackprof(void)
 #undef S
 
     gc_hook = Data_Wrap_Struct(rb_cObject, stackprof_gc_mark, NULL, &_stackprof);
+    _stackprof.fake_gc_frame = rb_str_new_literal("fake_gc_frame");
```
tmm1: You need to mark this as a global. It's getting GC'd otherwise.

ext/stackprof/stackprof.c
```diff
@@ -187,16 +188,24 @@ frame_i(st_data_t key, st_data_t val, st_data_t arg)
 
     rb_hash_aset(results, rb_obj_id(frame), details);
 
-    name = rb_profile_frame_full_label(frame);
-    rb_hash_aset(details, sym_name, name);
+    if (frame == _stackprof.fake_gc_frame) {
+	name = rb_str_new_literal("(garbage collection)");
+	file = rb_str_new_literal("");
```
tmm1: Might be worth it to create these two strings once upfront and mark them as globals to re-use as well.

jlfwong:
Feedback should be addressed! Thanks for the super fast turnaround time!

ext/stackprof/stackprof.c
```diff
@@ -550,8 +552,13 @@ Init_stackprof(void)
 #undef S
 
     gc_hook = Data_Wrap_Struct(rb_cObject, stackprof_gc_mark, NULL, &_stackprof);
+
     _stackprof.fake_gc_frame = rb_str_new_literal("fake_gc_frame");
-    rb_global_variable(&gc_hook);
```
tmm1: Did this get lost?

jlfwong: Indeed! Will fix.

jlfwong:
Still getting this on every 6th run or so, even after restoring the accidentally deleted line retaining `gc_hook` as a global variable.

```
........FFF.F/Users/jlfwong/code/stackprof/test/test_stackprof.rb:125: [BUG] Illegal instruction at 0x007fffd288ab44
```

Here's the c backtrace:

```
0   ruby                                0x000000010d814134 rb_vm_bugreport + 388
1   ruby                                0x000000010d6aed2a rb_bug_context + 490
2   ruby                                0x000000010d78ff38 sigsegv + 72
3   libsystem_platform.dylib            0x00007fffd2887bba _sigtramp + 26
4   stackprof.bundle                    0x000000010dd664d7 stackprof_results + 535
5   ruby                                0x000000010d8075be vm_call_cfunc + 302
6   ruby                                0x000000010d7f1181 vm_exec_core + 10049
7   ruby                                0x000000010d802114 vm_exec + 116
8   ruby                                0x000000010d80f5da invoke_block_from_c_splattable + 362
9   ruby                                0x000000010d7fd1fa rb_yield + 138
10  ruby                                0x000000010d64c9a9 rb_ary_each + 41
11  ruby                                0x000000010d8075be vm_call_cfunc + 302
12  ruby                                0x000000010d7f1181 vm_exec_core + 10049
13  ruby                                0x000000010d802114 vm_exec + 116
14  ruby                                0x000000010d80f5da invoke_block_from_c_splattable + 362
15  ruby                                0x000000010d7fd1fa rb_yield + 138
16  ruby                                0x000000010d65219a rb_ary_collect + 442
17  ruby                                0x000000010d8075be vm_call_cfunc + 302
18  ruby                                0x000000010d7f1181 vm_exec_core + 10049
19  ruby                                0x000000010d802114 vm_exec + 116
20  ruby                                0x000000010d800905 vm_invoke_proc + 293
21  ruby                                0x000000010d73d111 rb_proc_call + 81
22  ruby                                0x000000010d6b6f97 rb_exec_end_proc + 359
23  ruby                                0x000000010d6b74ce ruby_finalize_0 + 110
24  ruby                                0x000000010d6b75f5 ruby_cleanup + 277
25  ruby                                0x000000010d6b789d ruby_run_node + 61
26  ruby                                0x000000010d649faf main + 79
```

Another run that crashed yielded this stack trace:

```
0   ruby                                0x0000000101fd2134 rb_vm_bugreport + 388
1   ruby                                0x0000000101e6cd2a rb_bug_context + 490
2   ruby                                0x0000000101f4dee8 sigill + 72
3   libsystem_platform.dylib            0x00007fffd2887bba _sigtramp + 26
4   libsystem_platform.dylib            0x00007fffd288ab44 _os_unfair_lock_recursive_abort + 23
```

tmm1:
Getting the line number in stackprof_results where the crash is occurring would tell you what's going on. You'll want to compile the extension with CFLAGS="-ggdb -O0" and then run under lldb to catch the crash, then `bt` to see the backtrace with debug info.

jlfwong:
Okay, I got a little bit farther. It took me a while to figure out how to run the tests under lldb, but I got it eventually by using `ps aux | grep stackprof` while the tests were running via `bundle exec rake test`. The crashes happen intermittently which makes debugging frustrating.

It looks like malloc is failing somehow?

Here's the backtrace (at least for one of the classes of crashes).

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
  * frame #0: 0x00007fffd288ab44 libsystem_platform.dylib`_os_unfair_lock_recursive_abort + 23
    frame #1: 0x00007fffd288a465 libsystem_platform.dylib`_os_unfair_lock_lock_slow + 181
    frame #2: 0x00007fffd27fd445 libsystem_malloc.dylib`szone_malloc_should_clear + 179
    frame #3: 0x00007fffd27fd332 libsystem_malloc.dylib`malloc_zone_malloc + 107
    frame #4: 0x00007fffd27fc2b0 libsystem_malloc.dylib`malloc + 24
    frame #5: 0x00000001003b95f8 stackprof.bundle`stackprof_record_sample_for_stack + 88
    frame #6: 0x00007fffd2887bba libsystem_platform.dylib`_sigtramp + 26
    frame #7: 0x00007fffd27ff707 libsystem_malloc.dylib`tiny_free_list_add_ptr + 110
    frame #8: 0x00007fffd28144fc libsystem_malloc.dylib`tiny_free_no_lock + 406
    frame #9: 0x00007fffd28150f3 libsystem_malloc.dylib`free_tiny + 671
    frame #10: 0x000000010007bc1a ruby`ruby_xfree + 58
    frame #11: 0x00000001000885f4 ruby`obj_free + 740
    frame #12: 0x0000000100087d63 ruby`gc_sweep_step + 483
    frame #13: 0x000000010007fc7b ruby`garbage_collect + 139
    frame #14: 0x0000000100084603 ruby`gc_start_internal + 467
    frame #15: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #16: 0x00000001001a88dc ruby`vm_exec_core + 11932
    frame #17: 0x00000001001b9114 ruby`vm_exec + 116
    frame #18: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #19: 0x00000001001b414a ruby`rb_yield_1 + 74
    frame #20: 0x00000001000c7c58 ruby`int_dotimes + 216
    frame #21: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #22: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #23: 0x00000001001b9114 ruby`vm_exec + 116
    frame #24: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #25: 0x000000010006f9e4 ruby`rb_ensure + 164
    frame #26: 0x00000001003b9ed1 stackprof.bundle`stackprof_run + 65
    frame #27: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #28: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #29: 0x00000001001b9114 ruby`vm_exec + 116
    frame #30: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #31: 0x00000001001b414a ruby`rb_yield_1 + 74
    frame #32: 0x00000001000c7c58 ruby`int_dotimes + 216
    frame #33: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #34: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #35: 0x00000001001b9114 ruby`vm_exec + 116
    frame #36: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #37: 0x00000001001b41fa ruby`rb_yield + 138
    frame #38: 0x00000001000039a9 ruby`rb_ary_each + 41
    frame #39: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #40: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #41: 0x00000001001b9114 ruby`vm_exec + 116
    frame #42: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #43: 0x00000001001b41fa ruby`rb_yield + 138
    frame #44: 0x000000010000919a ruby`rb_ary_collect + 442
    frame #45: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #46: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #47: 0x00000001001b9114 ruby`vm_exec + 116
    frame #48: 0x00000001001b7905 ruby`vm_invoke_proc + 293
    frame #49: 0x00000001000f4111 ruby`rb_proc_call + 81
    frame #50: 0x000000010006df97 ruby`rb_exec_end_proc + 359
    frame #51: 0x000000010006e4ce ruby`ruby_finalize_0 + 110
    frame #52: 0x000000010006e5f5 ruby`ruby_cleanup + 277
    frame #53: 0x000000010006e89d ruby`ruby_run_node + 61
    frame #54: 0x0000000100000faf ruby`main + 79
    frame #55: 0x00007fffd267b255 libdyld.dylib`start + 1
```

I'm also not sure I set the debug flags correctly. To do that part, I made the following change:

```
diff --git i/ext/stackprof/extconf.rb w/ext/stackprof/extconf.rb
index 7d1a52c..ad29963 100644
--- i/ext/stackprof/extconf.rb
+++ w/ext/stackprof/extconf.rb
@@ -3,7 +3,10 @@ if have_func('rb_postponed_job_register_one') &&
    have_func('rb_profile_frames') &&
    have_func('rb_tracepoint_new') &&
    have_const('RUBY_INTERNAL_EVENT_NEWOBJ')
-  create_makefile('stackprof/stackprof')
+
+   with_cflags('-ggdb -O0') do
+     create_makefile('stackprof/stackprof')
+   end
 else
   fail 'missing API: are you using ruby 2.1+?'
 end
```

jlfwong:
It looks like there are at least 2 classes of failure. Here's a different one:

```
.Process 8145 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x107e5a000)
    frame #0: 0x00000001003ba4d7 stackprof.bundle`stackprof_results + 535
stackprof.bundle`stackprof_results:
->  0x1003ba4d7 <+535>: movq   (%rax,%r12), %rdi
    0x1003ba4db <+539>: addq   $0x8, %r12
    0x1003ba4df <+543>: decq   %r13
    0x1003ba4e2 <+546>: jne    0x1003ba4c0               ; <+512>
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x107e5a000)
  * frame #0: 0x00000001003ba4d7 stackprof.bundle`stackprof_results + 535
    frame #1: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #2: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #3: 0x00000001001b9114 ruby`vm_exec + 116
    frame #4: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #5: 0x00000001001b414a ruby`rb_yield_1 + 74
    frame #6: 0x00000001000c7c58 ruby`int_dotimes + 216
    frame #7: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #8: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #9: 0x00000001001b9114 ruby`vm_exec + 116
    frame #10: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #11: 0x00000001001b41fa ruby`rb_yield + 138
    frame #12: 0x00000001000039a9 ruby`rb_ary_each + 41
    frame #13: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #14: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #15: 0x00000001001b9114 ruby`vm_exec + 116
    frame #16: 0x00000001001c65da ruby`invoke_block_from_c_splattable + 362
    frame #17: 0x00000001001b41fa ruby`rb_yield + 138
    frame #18: 0x000000010000919a ruby`rb_ary_collect + 442
    frame #19: 0x00000001001be5be ruby`vm_call_cfunc + 302
    frame #20: 0x00000001001a8181 ruby`vm_exec_core + 10049
    frame #21: 0x00000001001b9114 ruby`vm_exec + 116
    frame #22: 0x00000001001b7905 ruby`vm_invoke_proc + 293
    frame #23: 0x00000001000f4111 ruby`rb_proc_call + 81
    frame #24: 0x000000010006df97 ruby`rb_exec_end_proc + 359
    frame #25: 0x000000010006e4ce ruby`ruby_finalize_0 + 110
    frame #26: 0x000000010006e5f5 ruby`ruby_cleanup + 277
    frame #27: 0x000000010006e89d ruby`ruby_run_node + 61
    frame #28: 0x0000000100000faf ruby`main + 79
    frame #29: 0x00007fffd267b255 libdyld.dylib`start + 1
```

tmm1:
That looks like classic signs of memory corruption, because some ruby object is not getting marked correctly.

ext/stackprof/stackprof.c
```diff
@@ -366,7 +366,7 @@ stackprof_record_sample_for_stack(int num)
 	    _stackprof.raw_samples = malloc(sizeof(VALUE) * _stackprof.raw_samples_capa);
 	}
 
-	if (_stackprof.raw_samples_capa <= _stackprof.raw_samples_len + num) {
+	while (_stackprof.raw_samples_capa <= _stackprof.raw_samples_len + (num + 2)) {
```
jlfwong: Okay! I think this fixes it.

I believe this code was previously unsafe.

This is checking to see if the capacity of `raw_samples` is sufficient to cover `num` new entries, but it really needs to cover up to `num + 2` new entries, since each sample entry is

stack height, followed by n frame ids, followed by the sample count.

I *think* this was an issue even before this change, but I'm not able to reproduce a crash before this PR.

Regardless, I think this is safer now?

tmm1: I'm not sure I follow.. why would this need to be a loop?

tmm1: I see, because we just double every time.

Calling realloc multiple times seems silly. It should calculate the required size once, and then call realloc once.

jlfwong: In practice, I suspect this will almost never need to call `realloc` more than once, because we set the initial buffer size to be 100 * the number of frames in the first sample, so for this to be called more than once, a given stack would have to be more than twice the size of all previous stacks combined. The `while` loop is just for safety, but I suspect in practice it'll never be a significant performance burden.

tmm1: Okay fair enough.

tmm1:
I see some random failures locally like:

```
  2) Failure:
StackProfTest#test_gc [/Users/tmm1/code/stackprof/test/test_stackprof.rb:132]:
Expected: 0
  Actual: 112
```

These should probably be attributed to the GC instead of counting as missed.

ext/stackprof/stackprof.c
```diff
@@ -411,26 +421,47 @@ stackprof_record_sample()
...
 stackprof_signal_handler(int sig, siginfo_t *sinfo, void *ucontext)
 {
     _stackprof.overall_signals++;
-    if (rb_during_gc())
-	_stackprof.during_gc++, _stackprof.overall_samples++;
-    else
-	rb_postponed_job_register_one(0, stackprof_job_handler, 0);
+    if (rb_during_gc()) {
+	_stackprof.during_gc++;
+	rb_postponed_job_register_one(0, stackprof_job_handler, (void*)1);
```
tmm1: Good catch. The list of signal-safe functions is very small, so calling into rb_hash_aset etc from the signal handler was not going to work reliably.

jlfwong:
> These should probably be attributed to the GC instead of counting as missed.
I guess it depends on what the miss count is meant to represent.
The miss count can be high there if the gc samples didn't have time to record.

It seems like you'd want to have the gc sample count to accurately reflect the number of sample in `raw` that correspond to garbage collection, and we can miss recording a sample if the next timer interrupt fires before the deferred jobs are run.

jlfwong:
I'm happy to make the sample counts correspond however you'd like -- just need to understand what `gc_samples` and `missed_samples` is meant to represent. Before this PR, it was pretty clear, but after this PR, the meaning of `gc_samples` seems a little murkier.

tmm1:
rb_postponed_job_register_one will not queue up more than one job for the same function callback. So when the signal fires multiple times during a GC, the job_handler fires only once and adds a single GC frame. This means the extra signals are considered "missed", even though we technically know they all happened during GC.

I think it makes sense to use distinct job handlers for gc and regular callbacks, so they don't cancel each other out.

And it seems we should use another counter like "current_gc_samples" that keeps track of how much time the current GC iteration took, then use that when the job finally fires and reset it at the end. Does that make sense?

jlfwong:
Yep, that makes sense! Thanks for the clear explanation

tmm1:
Diff looks great. Should I merge it?

jlfwong:
Well it looks like CI is failing? Do I need to address the failures there?

Seems like I'm violating some C90 restrictions?

jlfwong:
Looks like `rb_str_new_literal` doesn't exist in Ruby 2.1

tmm1:
That's probably worth fixing.

I just turned the CI on so I don't think anything other than 2.2 was passing before either.

tmm1:
> Looks like rb_str_new_literal doesn't exist in Ruby 2.1

You can detect if its available via extconf, and then add a wrapper using ifndef

jlfwong:
Well, that seems to have fixed the compilation failures. There are still test failures I'm not too sure what to do about. Some of them look like they're just differences in implementation, and some of them look like they probably randomly pass or succeed.

Is there anything you'd like me to do to address that? I think some of those problems predate this PR.

tmm1:
LGTM. I'm going to merge this. The obvious test failures due to method name being returned can probably be fixed with a `RUBY_VERSION` check in the test. But that can happen in a separate PR.

tmm1:
Tests are green on master. Thanks for working on this!

PR #89, opened 2017-11-18
Include timestamp deltas to get more accurate sample timings

### LINKED ISSUE #88: Include timestamps for each sample
Because we can't guarantee that samples happened at exactly the specified interval, it can be helpful to record timestamps for when each sample actually did occur.

This would allow us to generate sampling profiles that align correctly with wall-times (which can help diagnose where time is really going).

The Chrome sampling profile file format includes a list of timestamps -- one per sample.

Is this something you'd be interesting in accepting a PR for?

If so, I could use some guidance about what the right approach for collecting the timestamps is.

Based on https://stackoverflow.com/questions/12392278/measure-time-in-linux-time-vs-clock-vs-getrusage-vs-clock-gettime-vs-gettimeof, it seems like the safest, most portable way of getting walltime is `gettimeofday()`, though it sounds like there are better options if we have some `#if DARWIN` code in there. The downside of `gettimeofday()` is that it's not monotonic in cases of e.g. leap seconds. Given the purposes of stackprof, that seems like it should be fine?

I'm also interesting in suggestions for how to store the data. It could be stored as a stream of timestamps, or could be stored as a stream of time elapsed since the previous sample, since the absolute timestamps are roughly meaningless.

It also seems tricky to record when each GC sample was recorded, since we accumulate multiple GC samples to be recorded in a single deferred job execution, and we presumably don't want to allocate memory to store the timestamps of the GC samples.

jlfwong:
This includes a new field in the report output when `raw: true` is set.

The field `:raw_timestamp_deltas` indicates the amount of elapsed time in microseconds since the previous sample occurred. This allows us to match up wall time measurements with the sampling profiler.

Fixes #88

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -39,6 +39,11 @@ static struct {
     size_t raw_samples_capa;
     size_t raw_sample_index;
 
+    struct timeval last_sample_at;
+    int *raw_timestamp_deltas;
+    size_t raw_timestamp_deltas_len;
+    size_t raw_timestamp_deltas_capa;
+
     size_t overall_signals;
     size_t overall_samples;
     size_t during_gc;
@@ -54,7 +59,7 @@ static struct {
 
 static VALUE sym_object, sym_wall, sym_cpu, sym_custom, sym_name, sym_file, sym_line;
 static VALUE sym_samples, sym_total_samples, sym_missed_samples, sym_edges, sym_lines;
-static VALUE sym_version, sym_mode, sym_interval, sym_raw, sym_frames, sym_out, sym_aggregate;
+static VALUE sym_version, sym_mode, sym_interval, sym_raw, sym_frames, sym_out, sym_aggregate, sym_raw_timestamp_deltas;
 static VALUE sym_gc_samples, objtracer;
 static VALUE gc_hook;
 static VALUE rb_mStackProf;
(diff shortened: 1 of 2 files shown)


ext/stackprof/stackprof.c
```diff
@@ -428,24 +464,53 @@ stackprof_record_sample_for_stack(int num)
...
+    }
 }
 
 void
 stackprof_record_sample()
 {
+    int timestamp_delta = 0;
+    if (_stackprof.raw) {
+	struct timeval t;
+	gettimeofday(&t, NULL);
```
jlfwong: I looked at `Process.clock_gettime`, and it looks like it only has support for OS X >= 10.12 (and much older versions of linux) https://github.com/ruby/ruby/blob/fe3decb2017af2092d4b2181e27e4cd8f6bbf4fa/process.c#L7129

If that's fine with you, then I'm happy to update this PR to use that instead, and have these stored as doubles rather than as ints.

jlfwong:
I ran this on my local server that I'm profiling and was able to get the walltime timings to match up almost exactly! Exciting!

ext/stackprof/stackprof.c
tmm1: I don't have a strong preference. This is probably fine.

jlfwong: Okay, in that case this should be ready to review!

Issue #94, opened 2018-01-05
Filename is ignored in middleware mode; default filename is always used 


SherModo:
https://github.com/tmm1/stackprof/blob/master/lib/stackprof/middleware.rb#L27
https://github.com/tmm1/stackprof/blob/master/lib/stackprof/middleware.rb#L43
https://github.com/tmm1/stackprof/blob/master/lib/stackprof/middleware.rb#L46

SherModo:
I made these changes to make it work locally:
```
def initialize(app, options = {})
      @app       = app
      @options   = options
      @num_reqs  = options[:save_every] || nil

      Middleware.mode     = options[:mode] || :cpu
      Middleware.interval = options[:interval] || 1000
      Middleware.raw      = options[:raw] || false
      Middleware.enabled  = options[:enabled]
      Middleware.out      = options[:out] || nil
      Middleware.path     = 'tmp'
      at_exit{ Middleware.save } if options[:save_at_exit]
    end
```

```
def save
        if results = StackProf.results
          if Middleware.out
            filename = File.basename(Middleware.out)
            path = File.dirname(Middleware.out)
          else
            filename = "stackprof-#{results[:mode]}-#{Process.pid}-#{Time.now.to_i}.dump"
            path = Middleware.path
          end
          FileUtils.mkdir_p(path)
          File.open(File.join(Middleware.path, filename), 'wb') do |f|
            f.write Marshal.dump(results)
          end
          filename
        end
      end
```

tmm1:
Please submit a pull request.

itsderek23:
Fixed by #95.

PR #95, opened 2018-01-06
[#94] fix default use of filename in middleware and allow `out` option

SherModo:
Previously, it wasn't possible to pass filename in middleware call and make it override default. Also, the `out:` option available in Ruby wasn't available to the middleware. This PR fixes the former and adds the latter.

### DIFF
--- a/lib/stackprof/middleware.rb
+++ b/lib/stackprof/middleware.rb
@@ -11,7 +11,8 @@ module StackProf
       Middleware.interval = options[:interval] || 1000
       Middleware.raw      = options[:raw] || false
       Middleware.enabled  = options[:enabled]
-      Middleware.path     = options[:path] || 'tmp'
+      Middleware.out      = options[:out] || nil
+      Middleware.path     = 'tmp'
       at_exit{ Middleware.save } if options[:save_at_exit]
     end
 
@@ -30,7 +31,7 @@ module StackProf
     end
 
     class << self
-      attr_accessor :enabled, :mode, :interval, :raw, :path
+      attr_accessor :enabled, :mode, :interval, :raw, :path, :out
 
       def enabled?(env)
         if enabled.respond_to?(:call)
@@ -40,11 +41,17 @@ module StackProf
         end
       end
 
-      def save(filename = nil)
+      def save
         if results = StackProf.results
-          FileUtils.mkdir_p(Middleware.path)
-          filename ||= "stackprof-#{results[:mode]}-#{Process.pid}-#{Time.now.to_i}.dump"
-          File.open(File.join(Middleware.path, filename), 'wb') do |f|
+          if Middleware.out
+            filename = File.basename(Middleware.out)
+            path = File.dirname(Middleware.out)
+          else
+            filename = "stackprof-#{results[:mode]}-#{Process.pid}-#{Time.now.to_i}.dump"
+            path = Middleware.path
+          end
+          FileUtils.mkdir_p(path)
+          File.open(File.join(path, filename), 'wb') do |f|
             f.write Marshal.dump(results)
           end
           filename
--- a/lib/stackprof/middleware.rb
+++ b/lib/stackprof/middleware.rb
@@ -11,8 +11,8 @@ module StackProf
       Middleware.interval = options[:interval] || 1000
       Middleware.raw      = options[:raw] || false
       Middleware.enabled  = options[:enabled]
-      Middleware.out      = options[:out] || nil
-      Middleware.path     = 'tmp'
+      options[:path]      = 'tmp/' if options[:path].to_s.empty?
+      Middleware.path     = options[:path]
       at_exit{ Middleware.save } if options[:save_at_exit]
     end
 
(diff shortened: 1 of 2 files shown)


lib/stackprof/middleware.rb
```diff
@@ -11,7 +11,8 @@ def initialize(app, options = {})
       Middleware.interval = options[:interval] || 1000
       Middleware.raw      = options[:raw] || false
       Middleware.enabled  = options[:enabled]
-      Middleware.path     = options[:path] || 'tmp'
+      Middleware.out      = options[:out] || nil
+      Middleware.path     = 'tmp'
```
tmm1: Looks like the path parameter got lost?

SherModo: I intentionally did this because the path would be either extracted from the `out` option or default to `tmp`. If the user specifies a file only (no path, like in `out: 'my_file.dump'`), `File.dirname("my_file.dump")` will return `"."` which will cause the file to be stored the in the project's root directory (`./my_file.dump`).

SherModo: In other words, `tmp` will only be used if specified directly in the path passed to `out` or when the `out` option is not passed at all.

tmm1: Hmm I see. I'm worried about breaking backwards compatibility. If someone was using only path before, it would break after upgrading.

SherModo: Ah, that's a good point. The other reason that made me remove `path` was that I didn't want to keep two overlapping options, but now that you mention this, a decision needs to be made on whether to just modify the way `path` was being set, or to add `out` option alongside. I also noticed that there are no references to the middleware's `path` option in the README, so unless someone read the code of that file and decided to use the `path` option, they wouldn't know it exists.

SherModo: The `out` option is also consistent with the one available for Ruby in the `StackProf.run` method.

tmm1: How about we check if path is a directory, and if not we treat it as a full file name? Then we don't need two options.

SherModo: That works for me. Should I go ahead and make that change?

tmm1: Yes, please

lib/stackprof/middleware.rb
```diff
@@ -40,11 +41,20 @@ def enabled?(env)
...
       end
 
-      def save(filename = nil)
+      def save
         if results = StackProf.results
-          FileUtils.mkdir_p(Middleware.path)
-          filename ||= "stackprof-#{results[:mode]}-#{Process.pid}-#{Time.now.to_i}.dump"
-          File.open(File.join(Middleware.path, filename), 'wb') do |f|
+          path = Middleware.path
+          is_directory = path != path.chomp('/')
```
tmm1: This could still break for any users who were passing a directory without a trailing slash, right?

I was thinking you could use File.directory?

SherModo:
It turns out that this is trickier than I expected. Below is the reasoning for the latest changes:
- Ruby's `file?` and `directory?` return `true` only is the file/directory already exist
- These are the tests I did:
```
# try one at a time including an empty options hash
options = {
  # path: 'path/to/dir/'
  # path: 'path/to/dir' # 'dir' will be treated as file name here
  # path: 'path/to/dir/my_file.dump'
  path: ''
}

if options.empty?
  input = nil
else
  input = options[:path]
end

options[:path] = 'tmp/' if options[:path].to_s.empty?

path = options[:path]
is_directory = path != path.chomp('/')

if is_directory
  filename = 'auto_generated.dump'
else
  filename = File.basename(path)
  path = File.dirname(path)
end
p "Input: #{input}"
p "Path: #{path}"
p "File: #{filename}"
p "Joined: #{File.join(path, filename)}"
```

lib/stackprof/middleware.rb
SherModo: It turns out that this is trickier than I expected. Below is the reasoning for the latest changes:
- Ruby's `file?` and `directory?` return `true` only is the file/directory already exist
- These are the tests I did:
```
# try one at a time including an empty options hash
options = {
  # path: 'path/to/dir/'
  # path: 'path/to/dir' # 'dir' will be treated as file name here
  # path: 'path/to/dir/my_file.dump'
  path: ''
}

if options.empty?
  input = nil
else
  input = options[:path]
end

options[:path] = 'tmp/' if options[:path].to_s.empty?

path = options[:path]
is_directory = path != path.chomp('/')

if is_directory
  filename = 'auto_generated.dump'
else
  filename = File.basename(path)
  path = File.dirname(path)
end
p "Input: #{input}"
p "Path: #{path}"
p "File: #{filename}"
p "Joined: #{File.join(path, filename)}"
```

SherModo: - also, if the user passes `path:''`, I override this with `tmp/auto_generated.dump` because it wouldn't make sense (_I believe_).

tmm1: Okay makes sense. Thanks!

tmm1: Hmm it looks like tests are breaking because they expect the old behavior.

SherModo: I am going to give it a try. Will let the tests run through GitHub (Don't use `travis-ci` locally).

SherModo: I guess the last failure is unrelated?

SherModo:
Thanks!

PR #98, opened 2018-06-26
Eliminate one loop over the stack by using `overall_samples`

tenderlove:
Previously, the `already_accounted_in_total` member on the
`frame_data_t` struct was used to keep track of whether or not a
particular frame had been seen in a stack.  For example, a recursive
function is only supposed to count the frame once in the `total_samples`
member.

Instead of using the `already_accounted_in_total` as a flag, this commit
uses the `overall_samples` value as a unique identifier for the frame.
`overall_samples` will be different on every call to
`stackprof_record_sample_for_stack`, but the value will stay the same
for the duration of that function.  That means we can use the value to
indicate whether or not a particular frame has been "seen" and
we can eliminate one loop over the stack.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -20,7 +20,7 @@
 typedef struct {
     size_t total_samples;
     size_t caller_samples;
-    int already_accounted_in_total;
+    size_t seen_at_sample_number;
     st_table *edges;
     st_table *lines;
 } frame_data_t;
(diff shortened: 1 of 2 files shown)


itsderek23:
Thanks @tenderlove!

Released in [v0.2.12](https://github.com/tmm1/stackprof/blob/master/CHANGELOG.md#0212).

tenderlove:
Great, thank you!

PR #99, opened 2018-06-30
Remove /ext from .gitignore

brixen:
The /ext directory contains source code and ignoring this directory causes eg Atom to not render it in its tree view. Perhaps specific build artifacts should be ignored in place of the entire directory?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@
 /lib/stackprof/stackprof.bundle
 /lib/stackprof/stackprof.so
 *.sw?
-/ext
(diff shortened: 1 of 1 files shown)


tmm1:
Makes sense, since rake-compiler puts artifacts in ./tmp anyway

brixen:
Thanks @tmm1! I'm sure my PR isn't the complete solution, but it's confusing to clone the repo and not even see the `/ext` dir. There may be artifacts that end up there, but they should be explicitly excluded.

PR #102, opened 2018-10-02
Allow Pathname objects for Stackprof :out

### LINKED ISSUE #79: Using a Ruby Pathname as the output file results in error
In Ruby its common to use the helpful Pathname class to represent a path to a file. However, if you use this with stackprof you get an exception with no helpful stacktrace (because the IO flush happens in the compiled library).

```
2.4.0 :063 >  StackProf.run(mode: :cpu, out: Rails.root.join("tmp/stackprof-cpu.dump")) do
2.4.0 :064 >     10.times {2**10}
2.4.0 :065?>   end
NoMethodError: undefined method `flush' for nil:NilClass
	from (irb):63:in `run'

[rest of stacktrace removed because it's not helpful, and only shows the Ruby call stack down to the console]
```

However, if you call `.to_s`on the end of the pathname object then it works:
```
StackProf.run(mode: :cpu, out: Rails.root.join("tmp/stackprof-cpu.dump").to_s) do
2.4.0 :055 >     10.times {2**10}
2.4.0 :056?>   end
 => #<File:/[...]/tmp/stackprof-cpu.dump>
```

Presumably this is due to the fact that (upon reading the code) it seems that the `out` parameter can take eithert an IO object or a file path. In which case, this should be documented, and / or checked internally and converted to a string if necessary.

This caused much head-scratching and gnashing of teeth!

NickLaMuro:
The code for `stackprof_results` currently only does checks to see if the VALUE for `_stackprof.out` is a `T_STRING`, otherwise it assumes it is an IO object.

This updates the code to first check if `Pathname` has been required and defined.  If it is and the `_stackprof.out` is a `Pathname`, then open a file in the same manner as if it was a string, but calls `.to_s` to convert it to a ruby string first.

Fixes #79

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -312,7 +312,13 @@ stackprof_results(int argc, VALUE *argv, VALUE self)
 	if (RB_TYPE_P(_stackprof.out, T_STRING)) {
 	    file = rb_file_open_str(_stackprof.out, "w");
 	} else {
-	    file = rb_io_check_io(_stackprof.out);
+	    ID pathname_id = rb_intern("Pathname");
+	    if (rb_const_defined(rb_cObject, pathname_id) && 
+		    rb_obj_is_kind_of(_stackprof.out, rb_const_get(rb_cObject, pathname_id)) == Qtrue) {
+		file = rb_file_open_str(rb_funcall(_stackprof.out, rb_intern("to_s"), 0), "w");
+	    } else {
+		file = rb_io_check_io(_stackprof.out);
+	    }
 	}
 	rb_marshal_dump(results, file);
 	rb_io_flush(file);
--- a/test/test_stackprof.rb
+++ b/test/test_stackprof.rb
@@ -2,6 +2,7 @@ $:.unshift File.expand_path('../../lib', __FILE__)
 require 'stackprof'
 require 'minitest/autorun'
 require 'tempfile'
+require 'pathname'
 
 class StackProfTest < MiniTest::Test
   def test_info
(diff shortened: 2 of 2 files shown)


NickLaMuro:
### DISCLAIMER

This is my first time contributing a C extension on a ruby project, so I full expect this to be wrong on many fronts.  I invite you to mock, ridicule, tweet and laugh at with your friends, or completely scrap it and write it yourself in 1/20th of the time as you see fit.  🍅 's welcome.

NickLaMuro:
Also, regarding the test failure:

https://travis-ci.org/tmm1/stackprof/builds/435937186
https://travis-ci.org/tmm1/stackprof/jobs/435937189

Doesn't seem related to my code, but currently can't reproduce with ruby-2.3.7 locally, so it might be docker/travis specific in this case.  Will try re-pushing and to see if that fixes it, otherwise I will do a deeper dive into what is failing if that doesn't work.

tenderlove:
Hi @NickLaMuro.  Sorry it took so long for me to get to this.  I pulled in the test you wrote, but rewrote the patch to `stackprof.c`.  Basically I flipped the conditional for the `if` statement to check to see if the object responds to `to_io`.  If the object responds to `to_io` we'll assume it's an IO object we can write to, otherwise we'll treat it as a string.  In the "string" branch of the conditional pathname objects will Just Work, and the test that you added passes.

I don't understand what happened, but somehow when I was merging this in it removed your name from the commit on fd7d6fc6b6a26287a2ca71058a54b75427410c4d.  I'm really sorry about that.  I'd rewrite history to fix it, but I don't want to force push to master.  Again, I'm really sorry!

NickLaMuro:
@tenderlove No worries.  Wasn't really doing it for the "glory and riches", just saw the issue and noticed that it was unsolved for so long.  Happy we have a fixed patch in, and I will take a look at how you did it the #rightWay™ as well!

Thanks!

tenderlove:
No problem!  Thanks again for sending the PR, I really appreciate it! 😀

PR #103, opened 2018-10-26
Add --json format

palkan:
Add `--json` output to `stack prof` CLI.

The generated output could be used with web viewers (e.g. I want to use it with 
https://www.speedscope.app/ – yeah, I've noticed that there is a [PR](#100) to add out-of-the-box speedscope support; but who knows which flame graph viewer will rock in a year or two?).

### DIFF
--- a/lib/stackprof/report.rb
+++ b/lib/stackprof/report.rb
@@ -68,6 +68,11 @@ module StackProf
       f.puts Marshal.dump(@data.reject{|k,v| k == :files })
     end
 
+    def print_json(f=STDOUT)
+      require "json"
+      f.puts JSON.generate(@data)
+    end
+
     def print_stackcollapse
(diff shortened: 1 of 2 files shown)


lib/stackprof/report.rb
```diff
@@ -1,5 +1,6 @@
 require 'pp'
 require 'digest/md5'
+require 'json'
```
NickLaMuro: Could we move this into the `print_json` method so we are only requiring `JSON` when we are using it?

palkan: Yep; makes sense.

Done

jlfwong:
Sweet! I would love to get #100 merged, but this seems like a reasonable stop-gap to me!

I'm in an awkward position of technically having merge rights granted in this repository, but not rubygem publish rights, and also I'm not predominately a ruby dev.

@tenderlove any chance you have time to look at this and/or #100?

PR #107, opened 2019-02-09
bundle updated for ruby-head

walf443:
I sent #105  #106 but ruby-head spec was failed.

I'll try update Gemfile.lock to update bundler.


walf443:
Are there any reason to added Gemfile.lock to this project?
In many case, Gemfile.lock is used to ignored on ruby library project on my experience.

walf443:
hmm.

PR #110, opened 2019-05-09
Implement Report#merge for efficiently combining many reports

metcalf:
I need to combine 40 large profiles collected from a CI run. `Report#+` gets very slow as I merge more reports (e.g. with `reports.inject(:+)`). This PR implements `Report#merge` that allows for merging an arbitrary number of reports in a single pass. It also implements a test for the merge behavior.

### DIFF
--- a/lib/stackprof/report.rb
+++ b/lib/stackprof/report.rb
@@ -358,49 +358,66 @@ module StackProf
     end
 
     def +(other)
-      raise ArgumentError, "cannot combine #{other.class}" unless self.class == other.class
-      raise ArgumentError, "cannot combine #{modeline} with #{other.modeline}" unless modeline == other.modeline
-      raise ArgumentError, "cannot combine v#{version} with v#{other.version}" unless version == other.version
+      merge(other)
+    end
 
-      f1, f2 = normalized_frames, other.normalized_frames
-      frames = (f1.keys + f2.keys).uniq.inject(Hash.new) do |hash, id|
-        if f1[id].nil?
-          hash[id] = f2[id]
-        elsif f2[id]
-          hash[id] = f1[id]
-          hash[id][:total_samples] += f2[id][:total_samples]
-          hash[id][:samples] += f2[id][:samples]
-          if f2[id][:edges]
-            edges = hash[id][:edges] ||= {}
-            f2[id][:edges].each do |edge, weight|
-              edges[edge] ||= 0
-              edges[edge] += weight
-            end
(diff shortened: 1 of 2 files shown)


lib/stackprof/report.rb
```diff
@@ -358,49 +358,70 @@ def print_file(filter, f = STDOUT)
...
           end
-        else
-          hash[id] = f1[id]
         end
-        hash
       end
 
-      d1, d2 = data, other.data
-      data = {
+      all_data = all.map(&:data)
```
NickLaMuro: I feel like you could save a few iterations of this data by just calculating the `samples`, `gc_samples`, and `missed_samples` in a single loop:

```ruby
# untested code... you have been warned!
new_samples, new_gc_samples, new_missed_samples = all.inject([0,0,0]) do |result, data|
                                                    result[0] += data[:samples]
                                                    result[1] += data[:gc_samples]
                                                    result[2] += data[:missed_samples]
                                                    result
                                                  end

new_data = {
  # ...
  samples: new_samples,
  gc_samples: new_gc_samples,
  missed_samples: new_missed_samples
}
```

However, I realize it looks a bit uglier this way...

lib/stackprof/report.rb
```diff
@@ -358,49 +358,70 @@ def print_file(filter, f = STDOUT)
...
+      end
+
+      all = [self] + others
+      merged = {}
+
+      all.each do |report|
+        report.normalized_frames.each do |id, frame|
+          if !merged[id]
+            merged[id] = frame
+            next
```
NickLaMuro: I like the early eject here.  👍

lib/stackprof/report.rb
metcalf: I'm not sure it's worth optimizing this much since it's O(number of reports) whereas the main loop is O(frames) which should usually be many orders of magnitude larger.

metcalf:
@NickLaMuro thanks for the review!

metcalf:
As far as I can tell the test failures on recent Ruby versions are unrelated to this change.

lib/stackprof/report.rb
NickLaMuro: Yeah, also that.  Just something I noticed when looking through the code that stuck out to me, but definitely agree that it probably isn't the big focal point of the optimization.

I think it can be adjusted at a second pass if it turns out to be beneficial, but I think what you have is more readable and fine to keep as is.

PR #111, opened 2019-06-08
Fix a compilation error and a compilation warning with clang 6.0

Confusion:
Two occurrences of `_stackprof.interval` (a `VALUE`) were not converted into C longs in places they were added to, or an alternative to, a C long.


Confusion:
Build failures are due to issues with the build pipeline, not with the commit.

PR #112, opened 2019-09-16
Add `--alphabetical-flamegraph` for population-based instead of timeline

mame:
The current flamegraph of stackprof is often chopped because it puts the
passage of time on the x-axis.

This change adds `--alphabetical-flamegraph` mode which sorts the stack
traces in alphabetical order.  This view is much easier than the current
one for understanding bottlenecks.

Also, `--timeline-flamegraph` is added for consistency, and the original
`--flamegraph` option is the same effect as `--timeline-flamegraph` (for
compatibility reason).

### DIFF
--- a/bin/stackprof
+++ b/bin/stackprof
@@ -19,7 +19,9 @@ parser = OptionParser.new(ARGV) do |o|
   o.on('--graphviz', "Graphviz output (use with dot)"){ options[:format] = :graphviz }
   o.on('--node-fraction [frac]', OptionParser::DecimalNumeric, 'Drop nodes representing less than [frac] fraction of samples'){ |n| options[:node_fraction] = n }
   o.on('--stackcollapse', 'stackcollapse.pl compatible output (use with stackprof-flamegraph.pl)'){ options[:format] = :stackcollapse }
-  o.on('--flamegraph', "timeline-flamegraph output (js)"){ options[:format] = :flamegraph }
+  o.on('--timeline-flamegraph', "timeline-flamegraph output (js)"){ options[:format] = :timeline_flamegraph }
+  o.on('--alphabetical-flamegraph', "alphabetical-flamegraph output (js)"){ options[:format] = :alphabetical_flamegraph }
+  o.on('--flamegraph', "alias to --timeline-flamegraph"){ options[:format] = :timeline_flamegraph }
   o.on('--flamegraph-viewer [f.js]', String, "open html viewer for flamegraph output\n\n"){ |file|
     puts("open file://#{File.expand_path('../../lib/stackprof/flamegraph/viewer.html', __FILE__)}?data=#{File.expand_path(file)}")
     exit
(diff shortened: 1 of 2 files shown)


mame:
This is an example of `--alphabetical-flamegraph` of [a project of mine](https://github.com/mame/optcarrot):

image

This is the traditional `--flamegraph`:

image

For me, the alphabetical flamegraph is much easier to analyze the bottleneck.

---

BTW, [the original site of flame graph](http://www.brendangregg.com/flamegraphs.html) says:

> Summary
> The x-axis shows the stack profile population, sorted alphabetically (it is not the passage of time)

also says:

> Flame charts were first added by Google Chrome's WebKit Web Inspector (bug). While inspired by flame graphs, flame charts put the passage of time on the x-axis instead of the alphabet.

According to the article, the current `--flamegraph` of stackprof does not generate flamegraph, but flamechart.  So, it would be better to make `--flamegraph` an alias to `--alphabetical-flamegraph`.

NickLaMuro:
@mame Just an FYI:  What you described is exactly how the https://github.com/tmm1/stackprof/blob/master/bin/stackprof-flamegraph.pl works when using that (thought not well documented).

Brendan Gregg (the original creator of [that script](https://github.com/brendangregg/FlameGraph) ) talks a lot about about the specific difference between what he has dubbed "flamegraphs" (sorted alphabetically) and "flame charts" (sorted by time) and what you find typically in google's web inspector.  This project has unfortunately has put the two under the same umbrella (since forever ago), and conflated the two, along with projects like this one:

https://github.com/SamSaffron/flamegraph

So that doesn't help.  More info here:

http://www.brendangregg.com/flamegraphs.html

mame:
@NickLaMuro I didn't know the combination of `--stackcollapse` and `flamegraph.pl`.  It should be documented, but now I think `--d3-flamegraph` (#113) is more useful for me than `--alphabetical-flamegraph` and `flamegraph.pl`.  So, it looks good to me just to mention `--d3-flamegraph`.  (Sorry for shameless plug!)

I understand the unfortunate confusion of the wording.  I have no opinion (nor enough experience) for which word we should pick up, so it is okay for me to keep `--flamegraph` for the timeline-based one.

Thank you,

NickLaMuro:
> Thank you,

No problem.

> (Sorry for shameless plug!)

Heh, no need to apologize, and if anything I said gave the impression that what you provided in this PR was "wrong", that was not the intent.  Just trying to share something that already existed and provides the same functionality as what you were adding, and is also showing it is a pretty common output format that is found in other tools like [rbspy](https://github.com/rbspy/rbspy).

I personally gravitate towards output formats that I can use across multiple languages and are pretty heavily battle tested like `flamegraph.pl`, but that is just a personal take.  That said, I also think both formats (the "stackcollapsed form", and the timeline based) each provide their own pros and cons, so there is definitely merit to having a viewer that can display both formats, and that is a definitely deficiency of `flamegraph.pl` viewer as it can't do both (I think...).

eregon:
> That said, I also think both formats (the "stackcollapsed form", and the timeline based) each provide their own pros and cons, so there is definitely merit to having a viewer that can display both formats, and that is a definitely deficiency of `flamegraph.pl` viewer as it can't do both (I think...).

It can actually, since https://github.com/brendangregg/FlameGraph/pull/179

NickLaMuro:
> It can actually, since brendangregg/FlameGraph#179

@eregon Well, if our version of `flamegraph.pl` wasn't from 6 years ago, I would agree:

https://github.com/tmm1/stackprof/tree/master/vendor/FlameGraph

😄 

But cool, that is good to know!

PR #113, opened 2019-09-16
Add `--d3-flamegraph` to output html using d3-flame-graph

mame:
This change adds a new flamegraph output based on [d3-flame-graph](https://github.com/spiermar/d3-flame-graph).

d3-flame-graph allows us to investigate a flamegraph interactively by
zooming in/out and searching.


mame:
An example output: https://optcarrot-stackprof-d3-flame-graph.netlify.com/

PR #115, opened 2019-09-19
Avoid JSON::NestingError when processing deep stacks

osyoyu:
[`JSON.generate`'s default depth limit](https://ruby-doc.org/stdlib-2.6.4/libdoc/json/rdoc/JSON.html#method-i-generate) is 100, and this limitation leads to a JSON::NestingError when processing relatively deep stacks with `--json` and `--d3-flamegraph`.

This PR works around this by simply passing `max_nesting: false` to JSON.generate.

### DIFF
--- a/lib/stackprof/report.rb
+++ b/lib/stackprof/report.rb
@@ -70,7 +70,7 @@ module StackProf
 
     def print_json(f=STDOUT)
       require "json"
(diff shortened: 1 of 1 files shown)


osyoyu:
Build failures seem to be unrelated to this commit 🙃

PR #117, opened 2019-10-02
Goodie bag of small changes

aried3r:
This PR contains some minor changes, each in a separate commit so I can easily remove them if they are not wanted.

Unrelated to this PR, but as for the CI, I saw #116, which looks very nice since it actually works :D, but is the added runtime for CI and complexity (debatable) something that's wanted? I get the frustration with ruby-head, but then again, it's a `HEAD` version — can't this be in the allowed-failures until 2.7 is officially released or Travis CI starts working again?

A compromise could also be to use a separate build step for ruby-head and use Travis CI for everything else. https://docs.travis-ci.com/user/build-stages/define-steps/

WDYT?


aried3r:
I'll have a look at the failed test runs. One is because of the updated rake dependency.

aried3r:
I fixed one failure, but one is remaining, [see here](https://travis-ci.org/tmm1/stackprof/jobs/592967317#L1391-L1393).

But I'm at a loss why this would only fail on Ruby 2.3. Any thoughts?

aried3r:
The error here (in docker) is the same as when running it on Travis directly (see #118) with Ruby 2.3.

#118: https://travis-ci.org/tmm1/stackprof/jobs/592988044#L290-L292
#117: https://travis-ci.org/tmm1/stackprof/jobs/592967317#L1391-L1393

TylerRick:
> WDYT?

Ooh, a goodie bag!

I think the owner of this repo should add you as a collaborator so you can help maintain it and merge some of these PRs...

PR #118, opened 2019-10-03
Ubuntu 18.04 on Travis, try to return to non-docker CI builds

aried3r:
This is just a proposal. I do like the solution offered by @mame in #116, but maybe we can switch back to Travis itself, having faster builds.

What do you think in general of this approach?

PS: I'm opening this as a draft PR, since I don't actually know if it's gonna solve the `ruby-head` problems. But since the distribution used currently is an EOL Ubuntu 14.04, I thought I might as well try. :)


aried3r:
Ok, seeing all this, I believe the Docker CI setup is not so bad after all 😅

aried3r:
Closing, seems I can only pick the errors I want by switching distros.

PR #120, opened 2019-10-10
Rake tasks for running tests in Docker

NickLaMuro:
This is a set of helper rake tasks are intended to make running tests locally in docker much streamline and simpler to pick up as a new-comer to the project.

Requires a `docker` environment available in your shell environment.

Quick Start
-----------

To simply run the rake tasks, you can run:

```console
$ rake test:docker
```

To specify a specific target ruby version:

```console
$ rake test:docker:2.1
```

@mame Curious about your thoughts on this one since you recently did the work https://github.com/tmm1/stackprof/pull/116 to start running tests on travis with `docker`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,2 @@
 .ruby-version
+Gemfile.lock
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.ruby-version
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ env:
     - RVM_RUBY_VERSION=ruby-head
 
 before_install:
+- sudo --help
 - sudo -E which ruby
 - sudo -E which rake
 - sudo -E rake --trace test:docker:$RVM_RUBY_VERSION:build
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 services:
   - docker
 
-language: general
+language: ruby
 
 env:
   matrix:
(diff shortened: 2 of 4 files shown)


NickLaMuro:
Okay... sorry for all of the "commit vomit", but I think I am mostly at a good spot with this now.

As mentioned in the second-to-last commit, this is mostly where I think I will leave the final solution, but wanted to leave the `[FIXUP]` and `[DEBUG]` commits for reviewers in case they were curious about the process.  I plan on rebasing most of them back into the original 4 commits.

Regarding some of the tests failures, it seems like there is still a sporadic test failure that I have see happen multiple times now:

- https://travis-ci.org/tmm1/stackprof/jobs/595915372#L1139
- https://travis-ci.org/tmm1/stackprof/jobs/595911600#L1343
- https://travis-ci.org/tmm1/stackprof/jobs/595864466#L1393

The last one in that list is from my other PR, #119 (which that failure prompted this PR), so it isn't related to any of my changes (I think).  Probably can be fixed in a different PR, but possibly just raising the threshold to 20 will make it behave better on Travis...

¯\\(°\_o)/¯

NickLaMuro:
@tmm1 (or anyone for that matter):  Is this something of interest that I should keep open or should I close.

Had a few bugs from when I last left it with in CI that I had to resolve, and would of course rebase prior to a merge, but if this is adding complexity you would rather not have then I don't mind just closing.

Thanks.

NickLaMuro:
Now not relevant with the recent changes to use GitHub Actions.  Closing.

Issue #123, opened 2019-11-06
Program hangs if sample-collection time is > interval


benbuckman:
### Context
We are trying to enable StackProf in production on a REST API service. We would like to use `mode: :wall, raw: true, aggregate: true`, but it has been challenging to calibrate the `interval`. The segfault bug described in #81 makes very low/fast sampling rates concerning (especially because our hypothesis about the causes of those segfaults, which we have not verified, centered on [overlapping signal handlers](http://man7.org/linux/man-pages/man7/signal-safety.7.html)). But at an interval of (e.g.) 20ms the data is not particularly useful.

### Observed behavior
At lower intervals, however, we encounter a separate bug in which the program doesn't crash, but rather it _hangs_ – the CPU is pegged to 100% and the program being profiled does not advance. This specifically seems to occur **when the time to capture a single sample is greater than the interval between samples.** (The time it takes to capture a sample fluctuates, I assume, based on the size of the stack, CPU load, etc, so it could take a bunch of samples successfully and then encounter this race.) What I suspect is occurring – and I am still in the process of debugging and verifying the details – is that it starts [piling up jobs with `rb_postponed_job_register_one`](https://github.com/tmm1/stackprof/blob/b886c3c/ext/stackprof/stackprof.c#L561-L570), and by the time it finishes one sample, it's started another, (and so on forever), blocking any other instructions.

### Proposal
I think some code needs to be added, to [`stackprof_record_sample`](https://github.com/tmm1/stackprof/blob/b886c3c/ext/stackprof/stackprof.c#L491) or [`stackprof_record_sample_for_stack`](https://github.com/tmm1/stackprof/blob/b886c3c/ext/stackprof/stackprof.c#L381), measuring the time it took to capture the sample. If that time is `>= interval`, it needs to handle that somehow. I can think of a few options at this point:

1. Calculate how many sample(s)' time was borrowed and skip that number of next samples. (This seems complicated and risky.)
1. Inject fake frames into the results, similar to GC frames, that make it clear (to the final report or flamegraph) that this situation was encountered. (I'm not sure how feasible this would be.)
1. Internally stop profiling, and when `StackProf.results` is called, raise an exception that makes it clear that the profiling was stopped due to this situation. (We don't want to raise mid-profiling because that could bubble up anywhere in the program. This seems the easiest of the 3.)

### Next steps for me
- [ ] Distill this bug down to a simple repro script.
[...]

PR #124, opened 2019-11-07
Prevent hanging when sample collection time > interval

benbuckman:
This attempts to fix the bug described in #123, in which, when `interval` is faster than the time required to record and analyze stack frames, the job queue piles up faster than it's flushed and the program hangs.

Pieces:
- Moves the `in_signal_handler` lock to the global `_stackprof` struct, and adds a check, and escape hatch, in `stackprof_signal_handler`.
- Adds a test which demonstrates the hanging behavior and the fix for it.
- Adds helpers `timeval_to_usec` and `diff_timevals_usec` for working with time durations more easily as `long int`s of microseconds, rather than `timeval` structs.
- Adds a `debug` mode with stdout+stderr output recording the internal timings and flow.
- Adds some comments and renames some variables for clarity.

More context on the underlying bug, and consideration of alternative approaches, is at https://github.com/tmm1/stackprof/issues/123.

Motivation: 
We would like to use StackProf on a large-scale Ruby application in production. The inability to set an interval that is both fast enough to produce useful data, and guaranteed to never hang the program, is a blocker to doing so.

Caveats:
- I am an amateur with C. It took a lot of trial and error with this code and https://repl.it/languages/c to get this working.
- I have not yet analyzed what the escape hatch in `stackprof_signal_handler` does to the quality of the final results. (At the end of the day, it's a kind of user or program error if the interval is so fast that the escape hatch is necessary. The question is how to deal with it.)
- I have mostly tested this with `wall` mode; I have not yet analyzed how this bug or fix work with `cpu` or `object` modes (though I suspect they're similar).

cc @NickLaMuro, thank you for your engagement with #123!

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -129,9 +129,14 @@ stackprof_start(int argc, VALUE *argv, VALUE self)
 	sigaction(mode == sym_wall ? SIGALRM : SIGPROF, &sa, NULL);
 
 	timer.it_interval.tv_sec = 0;
-	timer.it_interval.tv_usec = NUM2LONG(interval);
+	timer.it_interval.tv_usec = NUM2INT(interval);
 	timer.it_value = timer.it_interval;
 	setitimer(mode == sym_wall ? ITIMER_REAL : ITIMER_PROF, &timer, 0);
+
+        if (_stackprof.debug) {
+            printf("started with interval %d (%ld sec %d usec)\n",
+                NUM2INT(interval), timer.it_interval.tv_sec, timer.it_interval.tv_usec);
+        }
     } else if (mode == sym_custom) {
 	/* sampled manually */
 	interval = Qnil;
@@ -396,7 +401,7 @@ st_numtable_increment(st_table *table, st_data_t key, size_t increment)
     captured `timestamp_delta` microseconds after previous sample.
 */
 void
-stackprof_record_sample_for_stack(int frame_count, int timestamp_delta)
+stackprof_record_sample_for_stack(int frame_count, int64_t timestamp_delta)
 {
     int i, n;
     VALUE prev_frame = Qnil;
@@ -514,10 +519,8 @@ stackprof_record_sample()
 
     if (_stackprof.debug) {
         int64_t time_since_start_usec = diff_timevals_usec(&_stackprof.started_at, &sampling_start);
-        printf("timestamp delta %ld usec since last, %ld since start, with interval %ld\n",
-            time_since_last_sample_usec,
-            time_since_start_usec,
-            NUM2LONG(_stackprof.interval));
+        printf("timestamp delta %lld usec since last, %lld since start, with interval %d\n",
+            time_since_last_sample_usec, time_since_start_usec, NUM2INT(_stackprof.interval));
     }
 
     frame_count = rb_profile_frames(0, sizeof(_stackprof.frames_buffer) / sizeof(VALUE), _stackprof.frames_buffer, _stackprof.lines_buffer);
@@ -529,14 +532,12 @@ stackprof_record_sample()
 
     if (_stackprof.debug) {
         int64_t sampling_duration_usec = diff_timevals_usec(&sampling_start, &sampling_finish);
-        printf("duration of stackprof_record_sample: %ld usec with interval %d\n",
-            sampling_duration_usec,
-            NUM2LONG(_stackprof.interval));
+        printf("duration of stackprof_record_sample: %lld usec with interval %d\n",
+            sampling_duration_usec, NUM2INT(_stackprof.interval));
 
-        if (sampling_duration_usec >= NUM2LONG(_stackprof.interval)) {
-            fprintf(stderr, "INTERVAL IS TOO FAST: %d with interval %d\n",
-                sampling_duration_usec,
-                NUM2LONG(_stackprof.interval));
+        if (sampling_duration_usec >= NUM2INT(_stackprof.interval)) {
+            fprintf(stderr, "INTERVAL IS TOO FAST: %lld with interval %d\n",
+                sampling_duration_usec, NUM2INT(_stackprof.interval));
         }
     }
 }
@@ -544,7 +545,7 @@ stackprof_record_sample()
 void
 stackprof_record_gc_samples()
 {
-    int delta_to_first_unrecorded_gc_sample = 0;
+    int64_t delta_to_first_unrecorded_gc_sample = 0;
     int i;
     struct timeval t;
     struct timeval diff;
@@ -564,7 +565,7 @@ stackprof_record_gc_samples()
     _stackprof.lines_buffer[0] = 0;
 
     for (i = 0; i < _stackprof.unrecorded_gc_samples; i++) {
-	int timestamp_delta = i == 0 ? delta_to_first_unrecorded_gc_sample : NUM2LONG(_stackprof.interval);
+	int timestamp_delta = i == 0 ? delta_to_first_unrecorded_gc_sample : NUM2INT(_stackprof.interval);
 	stackprof_record_sample_for_stack(1, timestamp_delta);
     }
     _stackprof.during_gc += _stackprof.unrecorded_gc_samples;
@@ -619,7 +620,7 @@ static void
 stackprof_newobj_handler(VALUE tpval, void *data)
 {
     _stackprof.overall_signals++;
-    if (RTEST(_stackprof.interval) && _stackprof.overall_signals % NUM2LONG(_stackprof.interval))
+    if (RTEST(_stackprof.interval) && _stackprof.overall_signals % NUM2INT(_stackprof.interval))
 	return;
     stackprof_job_handler(0);
 }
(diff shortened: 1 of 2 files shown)


ext/stackprof/stackprof.c
```diff
@@ -377,8 +387,13 @@ st_numtable_increment(st_table *table, st_data_t key, size_t increment)
     st_update(table, key, numtable_increment_callback, (st_data_t)increment);
 }
 
+/*
+    Records information about the frames captured in `_stackprof.frames_buffer`,
+    up to `frame_count-1` (buffer may contain more frames from prior sample),
+    captured `timestamp_delta` microseconds after previous sample.
+*/
```
benbuckman: (I'm not sure what the canonical comment format should be here.)

ext/stackprof/stackprof.c
```diff
@@ -482,41 +497,61 @@ stackprof_record_sample_for_stack(int num, int timestamp_delta)
 	prev_frame = frame;
     }
 
-    if (_stackprof.raw) {
-	gettimeofday(&_stackprof.last_sample_at, NULL);
-    }
+    gettimeofday(&_stackprof.last_sample_at, NULL);
```
benbuckman: (It seems simpler, and trivial, to assign this always, rather than track the downstream paths that do or don't need it.)

ext/stackprof/stackprof.c
```diff
@@ -482,41 +497,61 @@ stackprof_record_sample_for_stack(int num, int timestamp_delta)
...
-	struct timeval t;
-	struct timeval diff;
-	gettimeofday(&t, NULL);
-	timersub(&t, &_stackprof.last_sample_at, &diff);
-	timestamp_delta = (1000 * diff.tv_sec) + diff.tv_usec;
+    int frame_count;
+    struct timeval sampling_start, sampling_finish;
+
+    gettimeofday(&sampling_start, NULL);
+    long int time_since_last_sample_usec = diff_timevals_usec(&_stackprof.last_sample_at, &sampling_start);
```
benbuckman: (Note, this replaces the timestamp math fix in #122.)

ext/stackprof/stackprof.c
```diff
@@ -639,6 +682,22 @@ stackprof_atfork_child(void)
     stackprof_stop(rb_mStackProf);
 }
 
+long int timeval_to_usec(struct timeval *diff) {
+    return MICROSECONDS_IN_SECOND * diff->tv_sec + diff->tv_usec;
+}
+
+long int diff_timevals_usec(struct timeval *start, struct timeval *end) {
```
benbuckman: (I adapted this from an algorithm at https://www.guyrutenberg.com/2007/09/22/profiling-code-using-clock_gettime/, and tested it with this repl: https://repl.it/@BenBuckman/SanityCheck-Timevals)

ext/stackprof/stackprof.c
```diff
@@ -536,31 +571,39 @@ stackprof_record_gc_samples()
...
 {
     _stackprof.overall_signals++;
+
+    // Protect against individual samples taking longer to capture than
+    // the interval between samples, which would cause the job queue
+    // to pile up faster than it's flushed, peg the CPU, and hang the program.
+    if (_stackprof.in_signal_handler) {
+        if (_stackprof.debug)
+            fprintf(stderr, "skip stackprof_signal_handler, already in handler!\n");
+        return;
```
benbuckman: Another action we could consider doing in this case, is injecting fake frames, similar to GC, named `<skip>` or similar, so that the final report makes it clear that the sampling rate was too fast, rather than being silently lossy.

ext/stackprof/stackprof.c
tmm1: Instead of long int, include stdint.h and use int64_t

benbuckman: 2d3e274407a3b6fbc341cb67d291e39ea1e173b4

benbuckman:
@tmm1 Thanks for your feedback so far.
What do you think of this overall?
How would you like StackProf to handle the scenario described here – should it skip the signal handler (as currently implemented), or fill in fake frames, or stop profiling and raise a (e.g.) `IntervalTooFast` error later on `StackProf.results` (or at the end of `run`)?

benbuckman:
Thinking about it some more, another option would be to track and expose `_stackprof.skipped_frames`, and let the program use it after collecting the data, if it cares to.

benbuckman:
> `_stackprof.skipped_frames`

I see an integer `missed_samples` in the results,
defined at  https://github.com/tmm1/stackprof/blob/1d5ae02aade28447aec1915b65ba6a9d1598e8f4/ext/stackprof/stackprof.c#L655,
but it's never set in the C code, so it's always 0.
Was the original intention for `missed_samples` the same as I'm proposing here?

benbuckman:
@tmm1 || @tenderlove || other maintainers, do you have time to review this?
Thank you!

benbuckman:
@tmm1 I'm using `int64_t` now as suggested, but it's inconsistent – `_stackprof.raw_timestamp_deltas` still holds `int`s – and attempting to change that, like so

```diff
diff --git ext/stackprof/stackprof.c ext/stackprof/stackprof.c
index 458c9d5..6d5b18c 100644
--- ext/stackprof/stackprof.c
+++ ext/stackprof/stackprof.c
@@ -45,7 +45,7 @@ static struct {

     struct timeval started_at;
     struct timeval last_sample_at;
-    int *raw_timestamp_deltas;
+    int64_t *raw_timestamp_deltas;
     size_t raw_timestamp_deltas_len;
     size_t raw_timestamp_deltas_capa;

@@ -309,7 +309,7 @@ stackprof_results(int argc, VALUE *argv, VALUE self)
        raw_timestamp_deltas = rb_ary_new_capa(_stackprof.raw_timestamp_deltas_len);

        for (n = 0; n < _stackprof.raw_timestamp_deltas_len; n++) {
-           rb_ary_push(raw_timestamp_deltas, INT2FIX(_stackprof.raw_timestamp_deltas[n]));
+           rb_ary_push(raw_timestamp_deltas, LONG2FIX(_stackprof.raw_timestamp_deltas[n]));
        }
```

causes segfaults in tests (https://travis-ci.org/tmm1/stackprof/builds/610510787).

The remaining warnings are,
```
../../../../ext/stackprof/stackprof.c:476:75: warning: implicit conversion loses integer precision: 'int64_t' (aka 'long long') to 'int' [-Wshorten-64-to-32]
        _stackprof.raw_timestamp_deltas[_stackprof.raw_timestamp_deltas_len++] = timestamp_delta;
                                                                               ~ ^~~~~~~~~~~~~~~
../../../../ext/stackprof/stackprof.c:567:19: warning: comparison of integers of different signs: 'int' and 'size_t' (aka 'unsigned long') [-Wsign-compare]
    for (i = 0; i < _stackprof.unrecorded_gc_samples; i++) {
                ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../../ext/stackprof/stackprof.c:568:33: warning: implicit conversion loses integer precision: 'int64_t' (aka 'long long') to 'int' [-Wshorten-64-to-32]
        int timestamp_delta = i == 0 ? delta_to_first_unrecorded_gc_sample : NUM2INT(_stackprof.interval);
            ~~~~~~~~~~~~~~~            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Any idea why?

Separately, the test I added for the hanging-fix is failing in CI, but it's passing for me locally. The test is probabilistic, which is not great, but I don't see how else to test this. I'll try to at least make it pass consistently, and open to suggestions. 

Thanks

ext/stackprof/stackprof.c
```diff
@@ -113,9 +129,14 @@ stackprof_start(int argc, VALUE *argv, VALUE self)
...
 
 	timer.it_interval.tv_sec = 0;
-	timer.it_interval.tv_usec = NUM2LONG(interval);
+	timer.it_interval.tv_usec = NUM2INT(interval);
 	timer.it_value = timer.it_interval;
 	setitimer(mode == sym_wall ? ITIMER_REAL : ITIMER_PROF, &timer, 0);
+
+        if (_stackprof.debug) {
+            printf("started with interval %d (%ld sec %d usec)\n",
+                NUM2INT(interval), timer.it_interval.tv_sec, timer.it_interval.tv_usec);
```
benbuckman: (This highlights the separate issue of https://github.com/tmm1/stackprof/issues/125)

ext/stackprof/stackprof.c
```diff
@@ -130,8 +146,9 @@ stackprof_start(int argc, VALUE *argv, VALUE self)
     _stackprof.interval = interval;
     _stackprof.out = out;
 
-    if (raw) {
-	gettimeofday(&_stackprof.last_sample_at, NULL);
+    if (_stackprof.debug) {
+        printf("started with interval %d (%ld sec %d usec)\n",
```
benbuckman: (This highlights the separate issue of https://github.com/tmm1/stackprof/issues/125)

PR #128, opened 2019-12-04
Support custom metadata option in stackprof

imjching-shopify:
At the moment, there is lack of information about what we're profiling in the profiles themselves. This PR adds the ability to include custom metadata to each profile.

Note that the metadata for the profile can only be set through `stackprof_start` or `stackprof_run`. In the future, maybe we could include the ability to add more metadata when dumping the results through `stackprof_results`. I would imagine it to take in a custom hash, which will be merged with the hash that was supplied through `stackprof_start` or `stackprof_run`.

cc: @csfrancis @tenderlove @tmm1

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -33,6 +33,7 @@ static struct {
     VALUE mode;
     VALUE interval;
     VALUE out;
+    VALUE metadata;
 
     VALUE *raw_samples;
     size_t raw_samples_len;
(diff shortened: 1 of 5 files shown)


imjching-shopify:
Build failures seem to be unrelated to this commit. Past few merged PRs have the same errors as well. :smile:

ext/stackprof/stackprof.c
```diff
@@ -85,6 +86,14 @@ stackprof_start(int argc, VALUE *argv, VALUE self)
 	interval = rb_hash_aref(opts, sym_interval);
 	out = rb_hash_aref(opts, sym_out);
 
+  VALUE metadata_val = rb_hash_aref(opts, sym_metadata);
```
csfrancis: Indentation is off here.

imjching-shopify: Done.

PR #131, opened 2019-12-11
Remove Gemfile.lock from git

orien:
This file is only used in the development of the gem. Although it has been included in the released gem package it has no effect on the released artifact or environments where the stackprof gem is run.

Each stackprof developer should be able to create a personal one of these via running `bundle install`.

🤞 Hopefully, this fixes the CI build.


orien:
Looks like there's one test failure on Ruby HEAD. The jobs on other Rubies passed.

At least the tests are executing now. I think this is a step in the right direction.

PR #136, opened 2020-05-15
Faster flamegraph generation

jhawthorn:
This PR makes flamegraph generation significantly faster. The main change is to create cursor objects pointing into the raw buffer instead of slicing arrays out of the raw buffer.

**Before**
```
jhawthorn@guardian:~/src/stackprof (master ) [ruby 2.8.0]
$ time ruby -Ilib bin/stackprof --flamegraph stackprof-test-boot.dump > /dev/null
31.52s user 0.24s system 99% cpu 31.891 total
```

**After**
```
jhawthorn@guardian:~/src/stackprof (faster_reporting ) [ruby 2.8.0]
$ time ruby -Ilib bin/stackprof --flamegraph stackprof-test-boot.dump > /dev/null
6.69s user 0.16s system 99% cpu 6.875 total
```

I found this using stackprof on its own reporting 😆 ♻️

``` ruby
# frozen_string_literal: true

require "stackprof"

report = StackProf::Report.new(Marshal.load(IO.binread(ARGV[0])))

output = File.open("/dev/null", "wb")
StackProf.run(mode: :wall, out: 'stackprof-stackprof.dump') do
  report.print_timeline_flamegraph(output)
end
```

### DIFF
--- a/lib/stackprof/report.rb
+++ b/lib/stackprof/report.rb
@@ -95,17 +95,50 @@ module StackProf
       print_flamegraph(f, skip_common, true)
     end
 
+    StackCursor = Struct.new(:raw, :idx, :length) do
+      def weight
+        @weight ||= raw[1 + idx + length]
+      end
+
+      def [](i)
+        if i >= length
+          nil
+        else
+          raw[1 + idx + i]
+        end
+      end
+
+      def <=>(other)
+        i = 0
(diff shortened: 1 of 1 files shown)


tmm1:
This is really clever. Great use of stackprof!

PR #138, opened 2020-07-05
Update README.md

marcandre:
Should read "1000ms" or "1s" but not "1ms"


tmm1:
IIRC interval is specified in microseconds. See https://github.com/tmm1/stackprof#sampling

marcandre:
My bad, sorry for the noise. 🤦

PR #143, opened 2020-08-03
Make more accessible the profiler from TruffleRuby on GraalVM

mc2dx:
Related to https://github.com/oracle/truffleruby/issues/2044

Hi there,
During experiments with TruffleRuby and, of course, with Stackprof gem, I think that it would be beneficial if we can show information about the profiler from TruffleRuby/GraalVM. It can help for a newbie and somebody how to discover Ruby/GraalVM environments.
I know that it is no big deal but nice to have a feature.
Please let me know what you think about it.

### DIFF
--- a/ext/stackprof/extconf.rb
+++ b/ext/stackprof/extconf.rb
@@ -1,5 +1,7 @@
 require 'mkmf'
-if have_func('rb_postponed_job_register_one') &&
+if RUBY_ENGINE == 'truffleruby'
+  fail "try truffleruby's profiler \nruby --experimental-options --cpusampler --cpusampler.Mode=roots --cpusampler.SampleInternal FILE.rb"
+elsif have_func('rb_postponed_job_register_one') &&
    have_func('rb_profile_frames') &&
    have_func('rb_tracepoint_new') &&
    have_const('RUBY_INTERNAL_EVENT_NEWOBJ')
--- a/lib/stackprof.rb
+++ b/lib/stackprof.rb
@@ -1,7 +1,7 @@
 require "stackprof/stackprof"
 
 module StackProf
-  VERSION = '0.2.15'
+  VERSION = '0.2.16'
 end
 
 StackProf.autoload :Report, "stackprof/report.rb"
--- a/stackprof.gemspec
+++ b/stackprof.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name = 'stackprof'
-  s.version = '0.2.15'
+  s.version = '0.2.16'
   s.homepage = 'http://github.com/tmm1/stackprof'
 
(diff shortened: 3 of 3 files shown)


ext/stackprof/extconf.rb
```diff
@@ -1,5 +1,7 @@
 require 'mkmf'
-if have_func('rb_postponed_job_register_one') &&
+if RUBY_ENGINE == 'truffleruby'
+  fail "try truffleruby's profiler \nruby --experimental-options --cpusampler --cpusampler.Mode=roots --cpusampler.SampleInternal FILE.rb"
```
NickLaMuro, line 3: Not being a `TruffleRuby` user myself, I assume that the reasoning for this PR is that `stackprof` is not support on that platform (JVM?).  Maybe include that as part of the string here:

```ruby
fail "stackprof not supported on this platform (#{RUBY_ENGINE})!\n" \
     "for truffleruby, consider trying out truffleruby's profiler:\n\n" \
     "  $ ruby --experimental-options --cpusampler --cpusampler.Mode=roots --cpusampler.SampleInternal FILE.rb"
```

NickLaMuro: Also, not sure the `extconf.rb` is the best place for this, though I don't have an alternative off hand to suggest.

lib/stackprof.rb
```diff
@@ -1,7 +1,7 @@
 require "stackprof/stackprof"
 
 module StackProf
-  VERSION = '0.2.15'
+  VERSION = '0.2.16'
```
NickLaMuro, line 4: Generally speaking, I _**think**_ the version bumps have been done outside feature PRs in this repo.  Not positive on that statement, but something to consider.

mc2dx: I "brought" this practice from a different repository, but you are right.

ext/stackprof/extconf.rb
mc2dx: Maybe `mkmf` can be a better place but it is a more complex thing.

mc2dx:
I closed the PR because I am not sure if "discover alternative profilers for TruffleRuby/GraalVM environments" can be achieved by these changes.

NickLaMuro:
@mc2dx Does this maybe make sense to just put in the _Requirements_ section of the `README.md`?

https://github.com/tmm1/stackprof#requirements

Maybe change the requirement to be "C Ruby 2.2+", and add a note after along the lines of:

> For TruffleRuby/GraalVM Rubies, consider using the built in profilers that are available:
> 
> ```console
> $ ruby --experimental-options --cpusampler --cpusampler.Mode=roots --cpusampler.SampleInternal FILE.rb
> ```

mc2dx:
A small change but better than throwing exceptions.

PR #150, opened 2021-04-28
Collect stack frames immediately in Ruby 3.0

jhawthorn:
As of https://github.com/ruby/ruby/commit/0e276dc458f94d9d79a0f7c7669bde84abe80f21, which shipped in Ruby 3.0, it seems to be safe to collect stack frames inside the signal handler. This should allow more accurate results than waiting for the postponed job to run since that can only measure when interrupts are checked.

This new behaviour is wrapped inside a preprocessor check for Ruby 3+

Additionally, this moves the "in signal handler" checks up a level, and uses pthread_mutex_trylock, which should be more reliable and I _believe_ will fix the issue described in #123 and #124

@tenderlove and I had some discussion about whether this could have issues due to the writes in https://github.com/ruby/ruby/commit/0e276dc458f94d9d79a0f7c7669bde84abe80f21 being reordered.

One concern is whether there could be issues with hardware memory reordering (particularly on arm). I _believe_ this is safe only because we are only considering the stack from our current thread. These writes will appear consistent in our interrupt handler because of this.

Another concern is that the compiler could reorder the writes in https://github.com/ruby/ruby/commit/0e276dc458f94d9d79a0f7c7669bde84abe80f21. It doesn't seem to be, but that could absolutely happen. I think we should investigate making a change to Ruby to ensure the writes in `vm_push_frame`/`vm_pop_frame` aren't reordered.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -22,6 +22,14 @@
 #define FAKE_FRAME_MARK  INT2FIX(1)
 #define FAKE_FRAME_SWEEP INT2FIX(2)
 
+/*
+ * As of Ruby 3.0, it should be safe to read stack frames at any time
+ * See https://github.com/ruby/ruby/commit/0e276dc458f94d9d79a0f7c7669bde84abe80f21
+ */
+#if RUBY_API_VERSION_MAJOR < 3
+  #define USE_POSTPONED_JOB
+#endif
+
 static const char *fake_frame_cstrs[] = {
 	"(garbage collection)",
 	"(marking)",
@@ -630,7 +638,11 @@ stackprof_signal_handler(int sig, siginfo_t *sinfo, void *ucontext)
 	_stackprof.unrecorded_gc_samples++;
 	rb_postponed_job_register_one(0, stackprof_gc_job_handler, (void*)0);
     } else {
+#ifdef USE_POSTPONED_JOB
 	rb_postponed_job_register_one(0, stackprof_job_handler, (void*)0);
+#else
+	stackprof_job_handler(0);
+#endif
     }
     pthread_mutex_unlock(&lock);
 }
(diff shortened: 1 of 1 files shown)


eregon:
> One concern is whether there could be issues with hardware memory reordering (particularly on arm). I believe this is safe only because we are only considering the stack from our current thread. These writes will appear consistent in our interrupt handler because of this.

I think being in a signal handler gives up on the the "golden rule of multithreading": "a single thread should see everything it does as if it was done in that order".
My understanding is a signal handler is basically considered concurrency/like another thread in C.

AFAIK the signal could trigger anywhere, e.g. between two writes on `vm_push_frame` and that could cause issues, isn't it?

ivoanjo:
> AFAIK the signal could trigger anywhere, e.g. between two writes on vm_push_frame (even if they are in order, a big if) and that could cause issues, isn't it?

+1 I planned experimenting with this and raising it at some point -- the change in https://github.com/ruby/ruby/commit/0e276dc458f94d9d79a0f7c7669bde84abe80f21 did reorder things as far as the C source goes, but as far as I see it there really doesn't seem to be anything guaranteeing that the compiler won't reorder the write to `ec->cfp` with the actual initialization of the structure.

So... yeah this doesn't seem particularly safe at this moment.

(But it would be great if `rb_profile_frames` could indeed be made async-safe!)

mame:
As @eregon and @ivoanjo have said, I think there is a theoretical concern with this change.

However, it may not lead to a visible error on major architectures and C compilers. Even in a potentially reproducible environment, it will be very unlikely to be visible unless the signal handler is called at an extremely unlucky timing.

I think it is a possible option to ignore such a theoretical concern and take the practical benefits, but it's a difficult decision. I don't interfere with your decisions in stackprof, but if it were the interpreter itself, I would like to be very cautious.

PR #151, opened 2021-04-29
Use GitHub Actions for CI

casperisfine:
And modernize the Rakefile a bit to use the tasks provided by bundler etc.

Ruby 3.0 is failing just like on Travis. I'll see if I can figure it out.

cc @tenderlove @jhawthorn


casperisfine:
Hum, it doesn't compile on Windows,but I presume it's legit?

```
ext/stackprof/stackprof.c:146:31: error: 'SIGALRM' undeclared (first use in this function); did you mean 'SIGABRT'?
  146 |  sigaction(mode == sym_wall ? SIGALRM : SIGPROF, &sa, NULL);
      |                               ^~~~~~~
      |                               SIGABRT
```

IIRC Windows have limited signal support?

tenderlove:
Ya, I don't think we can use this on Windows (though maybe we should figure out a solution at some point)

PR #154, opened 2021-05-04
Convert StackProf.run / StackProf.result to Ruby

tenderlove:
This commit is kind of a yak shave.  I would like to add thread tracking
to stackprof.  I want to allow people to optionally dump the thread for
which the stack was collected on by doing `StackProf.result(threads: true)`
Updating the signature for `result` was kind of hard, so I am
refactoring it to be in Ruby.

This commit technically breaks the case of someone doing:

```ruby
StackProf.start(out: "some_file.data")
...
StackProf.result
```

But I think it's strange to ask the start function to dump the file
rather than asking the result function.  The result function actually
writes to the file, so I think it makes more sense to pass the file (or
file name) to the result function.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -53,7 +53,6 @@ static struct {
 
     VALUE mode;
     VALUE interval;
-    VALUE out;
     VALUE metadata;
     int ignore_gc;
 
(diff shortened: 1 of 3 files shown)


jhawthorn (approved):

tenderlove:
Since this is technically a breaking change, I'm going to bump the major version on master and remove some of the supported Ruby versions.

PR #161, opened 2021-07-06
Use monotonic clock for timestamp deltas when available

djudd:
Check for [_POSIX_MONOTONIC_CLOCK](https://linux.die.net/man/3/clock_gettime) and, if present, use `clock_gettime` with `CLOCK_MONOTONIC` instead of using `gettimeofday`.

I've verified that tests pass and use the monotonic clock on both Ubuntu 20.04.2 and macOS 11.4 (which apparently grew support for `clock_gettime` recently). Haven't tested on any other systems but did verify the tests still pass and _don't_ use the monotonic clock if I munge the check.

I don't write much C, so please tell me what terrible things I've done and how to make them better.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -13,10 +13,12 @@
 #include <ruby/intern.h>
 #include <signal.h>
 #include <sys/time.h>
+#include <time.h>
 #include <pthread.h>
 
 #define BUF_SIZE 2048
 #define MICROSECONDS_IN_SECOND 1000000
+#define NANOSECONDS_IN_SECOND 1000000000
 
 #define FAKE_FRAME_GC    INT2FIX(0)
 #define FAKE_FRAME_MARK  INT2FIX(1)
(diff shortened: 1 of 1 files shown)


ext/stackprof/stackprof.c
```diff
@@ -558,25 +592,23 @@ stackprof_record_sample()
 void
 stackprof_record_gc_samples()
 {
-    int delta_to_first_unrecorded_gc_sample = 0;
-    int i;
+    int64_t delta_to_first_unrecorded_gc_sample = 0;
+    size_t i;
```
djudd, line 596: This is an incidental change but made a compiler warning go away.

djudd:
Since I'm not sure @tmm1 is actively maintaining this anymore, @tenderlove want to take a look?

tenderlove:
It seems fine but I'll check more seriously (not from my phone lol) tomorrow. Thanks for sending this!! It makes sense to use the monotonic clock

djudd:
@tenderlove did you get a chance to check more seriously?

tenderlove:
@djudd sorry, I totally forgot.  Just reviewed now, and it looks great!

PR #162, opened 2021-08-26
Fix infinite realloc loop when rb_profile_frames returns 0

isobit:
Sometimes rb_profile_frames returns 0. If this happens before raw_samples is allocated for the first time, stackprof_record_sample_for_stack can enter an infinite loop as it tries to realloc raw_samples by doubling raw_samples_capa, which is 0 when num is 0.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -472,7 +472,7 @@ stackprof_record_sample_for_stack(int num, int64_t timestamp_delta)
 
     _stackprof.overall_samples++;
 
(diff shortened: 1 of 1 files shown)


isobit:
Looks like the tests ran into the same issue, submitted https://github.com/tmm1/stackprof/pull/164 to address it.

isobit:
Hey @tmm1 (or @tenderlove?), sorry again for the direct ping but could you take another look at this and consider merging, now that https://github.com/tmm1/stackprof/pull/164 is merged? Thanks!

PR #164, opened 2021-08-28
Make test_cputime more reliable

isobit:
It seems the order of the frames hash values in test_cputime is somewhat
nondeterministic, at least with Ruby > 3. Most of the time the
"StackProfTest#math" frame will be at index 1, but sometimes it shows up
at index 0, like it does with Ruby < 3. This commit updates the test to
accept it in either position.

### DIFF
--- a/test/test_stackprof.rb
+++ b/test/test_stackprof.rb
@@ -78,9 +78,14 @@ class StackProfTest < MiniTest::Test
     end
 
     assert_operator profile[:samples], :>=, 1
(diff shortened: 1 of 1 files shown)


isobit:
Sorry to bother you @tmm1 and @tenderlove but would you mind taking a look at this and https://github.com/tmm1/stackprof/pull/162? I wanted to make sure what I'm doing here makes sense to fix the inconsistent test failures I was seeing in that other PR.

tenderlove:
Yes, this makes sense to me.  Thank you!

PR #167, opened 2021-09-07
Fix timestamp math for platforms with _POSIX_MONOTONIC_CLOCK

ivoanjo:
Hello, hello! 🎉

First the disclaimer (I guess it may be relevant): I work at @DataDog on profiling for the [ddtrace](https://github.com/DataDog/dd-trace-rb) gem, although we don't use `rb_profile_frames` (but it is quite interesting).

I noticed that the fix for #122 in #163 actually missed the new `_POSIX_MONOTONIC_CLOCK` branch, where the math was still incorrect (used 1000 instead of 1000000 when attempting to convert seconds to microseconds).

cc @djudd @casperisfine

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -49,9 +49,9 @@ static const char *fake_frame_cstrs[] = {
   }
 
   static int64_t delta_usec(timestamp_t *start, timestamp_t *end) {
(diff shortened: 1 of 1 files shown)


casperisfine (approved):

tenderlove:
Thank you!

PR #170, opened 2021-11-08
Record sample timestamps in :raw mode (proposal)

djudd:
This PR adds a `:raw_sample_timestamps` key to results in `:raw` mode, which contains the (starting) timestamp of each sample in microseconds-since-some-arbitrary-time, which will be the monotonic clock zero on most operating systems or the epoch on systems that don't support it. This isn't quite redundant with the existing `:raw_timestamp_deltas` because the latter exclude time spent sampling in Stackprof - each subtracts the ending time of the previous sample from the starting time of this sample.

My use case is joining the Stackprof sample data against other tracing information which uses the same OS clock in post-processing. In particular, this is useful for working around the limitations of Stackprof in a multithreaded environment, where ownership of the GVL may return to an arbitrary thread while the main thread is blocked, and so you can no longer tell from the stack sample alone what the main thread was blocked on. However, if you have tracing around I/O calls plus these timestamps, you can reconstruct that information.

An alternate design would be to change the semantics of `:raw_timestamp_deltas` so that they no longer exclude Stackprof sample time, which would let you convert them to wall time with a cumulative sum. We could drop back to just capturing a timestamp once per sample. This might be how I'd have designed things from scratch but seems like a larger compatibility break.

Thoughts @tenderlove?

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -58,6 +58,10 @@ static const char *fake_frame_cstrs[] = {
       }
       return result;
   }
+
+  static uint64_t timestamp_usec(timestamp_t *ts) {
+      return (MICROSECONDS_IN_SECOND * ts->tv_sec) + (ts->tv_nsec / 1000);
+  }
 #else
   #define timestamp_t timeval
   typedef struct timestamp_t timestamp_t;
@@ -71,6 +75,10 @@ static const char *fake_frame_cstrs[] = {
       timersub(end, start, &diff);
       return (MICROSECONDS_IN_SECOND * diff.tv_sec) + diff.tv_usec;
   }
+
+  static uint64_t timestamp_usec(timestamp_t *ts) {
+      return (MICROSECONDS_IN_SECOND * ts.tv_sec) + diff.tv_usec
+  }
 #endif
 
 typedef struct {
@@ -80,6 +88,11 @@ typedef struct {
     st_table *edges;
     st_table *lines;
 } frame_data_t;
+
+typedef struct {
+    uint64_t timestamp_usec;
+    int64_t delta_usec;
+} sample_time_t;
 
 static struct {
     int running;
@@ -98,9 +111,9 @@ static struct {
     size_t raw_sample_index;
 
     struct timestamp_t last_sample_at;
-    int *raw_timestamp_deltas;
-    size_t raw_timestamp_deltas_len;
-    size_t raw_timestamp_deltas_capa;
+    sample_time_t *raw_sample_times;
+    size_t raw_sample_times_len;
+    size_t raw_sample_times_capa;
 
     size_t overall_signals;
     size_t overall_samples;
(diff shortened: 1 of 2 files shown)


ext/stackprof/stackprof.c
```diff
@@ -617,7 +638,7 @@ stackprof_record_gc_samples()
         _stackprof.lines_buffer[1] = 0;
         _stackprof.unrecorded_gc_marking_samples--;
 
-        stackprof_record_sample_for_stack(2, timestamp_delta);
+        stackprof_record_sample_for_stack(2, start_timestamp, timestamp_delta);
```
djudd, line 641: We end up using the same timestamp for each GC sample, which might look weird; maybe we should be spreading them out artificially the same way we do for time deltas. But that wasn't needed for my use case and this is simple.

test/test_stackprof.rb
```diff
@@ -126,19 +126,33 @@ def test_custom
...
+    assert_equal 10, profile[:raw_sample_timestamps].size
+    profile[:raw_sample_timestamps].each_cons(2) do |t1, t2|
+      assert_operator t1, :>, before_monotonic
+      assert_operator t2, :>=, t1
+      assert_operator t2, :<, after_monotonic
+    end
+
     assert_equal 10, profile[:raw_timestamp_deltas].size
+    total_duration = after_monotonic - before_monotonic
+    assert_operator profile[:raw_timestamp_deltas].sum, :<, total_duration
```
djudd, line 155: I wanted to also assert >0 but FWIW on both this branch and master I actually see negative time deltas on my machine (an M1 Macbook) in this test case. I assume this is only possible because we're sampling in a tight loop, have never seen it in prod, but it's weird and I couldn't figure out from the code how it could happen. I should probably file a separate bug for that.

djudd (commented):
FWIW I think this would let Stripe get back on to mainline Stackprof, we run something like this, but I'm reconstructing the core idea from memory since I'm on leave, and this actual patch hasn't been tested in prod.

tenderlove:
Just to clarify, it seems like the timestamp delta values could be calculated *from* the recorded timestamp samples (if we were to merge this PR).  Is that correct?

djudd:
I believe that is _not_ correct, unless we were to either change the semantics of the timestamp delta values, or separately track an array of duration-in-Stackprof-sampling values. The current timestamp delta values exclude time spent actually capturing the stack sample, since each is the diff between a timestamp captured on exit of Stackprof's capture code and on entry at the next sample. And this _can_ matter empirically - I spent a ~day trying to do the reverse (infer timestamps from deltas) before figuring out why my numbers wouldn't line up.

^ is what I was trying to get at with the paragraph in the PR description on "An alternate design... "

tenderlove:
Ah got it. I thought the deltas were calculated at the beginning for some reason. Regardless this seems fine so I'll merge it. Thanks!

PR #173, opened 2022-02-17
Update README

sydneybrokeit:
Examples use `out` instead of `path` for options.  This is invalid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Then run `$ bundle install`. Alternatively you can run `$ gem install stackprof`
 in ruby:
 
 ``` ruby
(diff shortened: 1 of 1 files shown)


chrisyuska:
Am I missing something here or should this be reverted?  Using `path` doesn't write any file and the tests still use `out`: https://github.com/tmm1/stackprof/blob/v0.2.19/test/test_stackprof.rb#L247-L283

casperisfine:
Yes, I just noticed the same thing, I thought I was going crazy...

chrisyuska:
I've submitted https://github.com/tmm1/stackprof/pull/176 to revert this documentation in case it was made by mistake.

sydneybrokeit:
This is the only way the middleware works.  If the main module and the
middleware require different options, that should be noted.

PR #174, opened 2022-02-24
Add json file support to profile reading

gmcgibbon:
Adds support for reading json serialized profile files so lines like `stackprof --d3-flamegraph profile.json > flamegraph.html` work.


lib/stackprof/report/parsing.rb
```diff
@@ -0,0 +1,39 @@
...
+module StackProf
+  class Report
+    module Parsing
+      def from_file(file)
+        content = IO.binread(file)
+        data = begin
+          from_json(JSON.parse(content))
+        rescue JSON::ParserError
+          new(Marshal.load(content))
+        end
```
tenderlove, line 12: Instead of trying to detect this via parsing / raising, can you check the first couple bytes?  [Rails does something like that](https://github.com/rails/rails/blob/40a910c9e9ed491443fcbcef43fab22ba73f0250/actionpack/lib/action_dispatch/middleware/cookies.rb#L556-L562) and I think it would be a little nicer.

tenderlove (changes requested):

lib/stackprof/report/parsing.rb
gmcgibbon: TIL. Thanks!

PR #186, opened 2022-08-22
Use postponed jobs on Ruby 2.x

casperisfine:
As pointed by @eregon, I made a mistake https://github.com/tmm1/stackprof/pull/180/files#r951294711

@tenderlove


eregon:
@ms-ati That's just a flakey test, it's not specific to macOS I've seen it fail randomly on different CI jobs.

eregon (approved):

PR #187, opened 2022-09-14
`stackprof run` CLI

casperisfine:
Sometime when you want to profile a ruby process for its entire lifetime, it's a bit complicated to find the proper entrypoint and exit point to insert `StackProf.start` etc.

This new CLI command allows to do this easily, e.g:

```bash
$ stackprof run -- rubocop --cache false
StackProf results dumped at: /var/folders/pg/ykz6j94s7dv_2z4l5x_m17l00000gn/T/stackprof20220914-26008-7iipjc.dump
```

Or for profiling a Rails application boot sequence:

```bash
$ stackprof run -- bin/rails runner ':ok'
```

The most common `StackProf.start` arguments can be passed as command line arguments as well.

NB: not too sure how this could be unit tested without major headaches :/.

FYI @eregon

### DIFF
--- a/bin/stackprof
+++ b/bin/stackprof
@@ -2,94 +2,129 @@
 require 'optparse'
 require 'stackprof'
 
-options = {}
+if ARGV.first == "run"
+  ARGV.shift
+  env = {}
+  parser = OptionParser.new(ARGV) do |o|
+    o.banner = "Usage: stackprof run [--mode|--out|--interval] -- COMMAND"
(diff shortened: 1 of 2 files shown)


eregon:
Nice. Also this seems like a mode we should be able to easily support for TruffleRuby, by passing --cpusampler to the truffleruby process, for CPU sampling with histogram and flamegraph outputs.

casperisfine:
cc @tenderlove @jhawthorn

PR #188, opened 2022-09-19
Correct parsing of opts for bin/rubocop

nate-at-gusto:
The mode and out options are not able to be used because they'll pass a `true` value into the ENV, which blows up.

* Improve helpstring
* Mode/out options on stackprof run can now be used


byroot (approved):
cc @tenderlove could you merge this?

PR #196, opened 2023-01-20
Use postponed jobs on Ruby 3.2.0

casperisfine:
Otherwise it can cause a VM crash.
Won't be a problem in 3.2.1.

cc @jhawthorn @tenderlove

### DIFF
--- a/lib/stackprof.rb
+++ b/lib/stackprof.rb
@@ -6,6 +6,11 @@ end
 
 if defined?(RubyVM::YJIT) && RubyVM::YJIT.enabled?
   StackProf.use_postponed_job!
+elsif RUBY_VERSION == "3.2.0"
+  # 3.2.0 crash is the signal is received at the wrong time.
+  # Fixed in https://github.com/ruby/ruby/pull/7116
+  # The fix is backported in 3.2.1: https://bugs.ruby-lang.org/issues/19336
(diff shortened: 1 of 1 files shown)


jhawthorn:
Can you elaborate on how this happens? As of 67ec8fe92c8c8578311be3d8f795a34003f5f79a we should be forwarding to the thread which stackprof was started from, I'd think that would sidestep the issue (we know that thread has a stack because we ran code to start stackprof).

casperisfine:
@jhawthorn I don't particularly understand the bug myself, all I know if that there is a race condition fixed by https://github.com/ruby/ruby/pull/7116. I'm sure @tenderlove or @XrXr can actually explain the issue.

tenderlove:
@jhawthorn I'm not 100% sure how it's happening.  I agree it's odd since we should be forwarding the signal to only active and live threads, but the core files we recovered definitely showed the ec was not fully set up.  Maybe there is a race condition when threads switch?

Regardless of the root cause, I think we should merge this.  At least it buys us time to figure out a stable repro.

ivoanjo:
@jhawthorn maybe the issue is in `mode: :cpu`, not `mode: :wall`?

byroot:
Yes, I forgot to mention that the crash only happens in CPU mode.

PR #199, opened 2023-02-27
Tiny structural improvement of README.

apotonick:
Add some headers an example for `mode: :object` in the very top.

Thanks for this tool, I am in love with it! :green_heart:


README.md
```diff
@@ -16,23 +16,38 @@ Inspired heavily by [gperftools](https://code.google.com/p/gperftools/), and wri
...
+#### Mode
+
+When profiling a less complex code chunk it might be more helpful to switch from `:cpu` mode to `:object`. This is
+advisable when your `:cpu` run doesn't produce any samples.
+
+``` ruby
+StackProf.run(mode: :object, out: "tmp/stackprof-cpu-myapp.dump") do
+  #...
+end
+```
```
tenderlove, line 46: I don't think this is accurate.  You should only switch from `:cpu` to `:object` when you want to change from profiling time to profiling object allocations.  If you run with `:cpu` doesn't produce any samples it means that your sampling frequency might not be high enough.  Measuring `:cpu` vs `:object` are completely different things (some code could take lots of time but allocate no objects, some code could run fast but allocate many objects).

tenderlove (changes requested):
Please update the docs to differentiate between cpu and object profiling.

Thanks!

Issue #157, opened 2021-05-28
Segmentation fault when stackprof is left running during VM shutdown


ivoanjo:
Howdy! 👋 

First the disclaimer (I guess it may be relevant): I work at @Datadog on profiling for the [ddtrace](https://github.com/DataDog/dd-trace-rb) gem, although we don't use `rb_profile_frames` (tradeoffs, tradeoffs 🤣 ).

While doing a few experiments with stackprof (0.2.17), I noticed that I got VM crashes when the process was finishing up with stackprof still enabled. I can reproduce this on both Linux and macOS, on Ruby 3.0.1, 2.7.3 and 2.6.7.

The following script is enough to reproduce it every time for me:

```ruby
require 'stackprof'

StackProf.start(mode: :wall, raw: true, aggregate: false, ignore_gc: true, interval: 100)
sleep 0.5
```

I went through a few core dumps (let me know if you'd like me to share them here) and the pattern seems that stackprof keeps sampling even through the VM is already "closing shop", in particular `GET_EC()` is `null` and so `rb_postponed_job_register_one` blows up as it tries to dereference it.

Here's one such example:

```
# ruby stackprof-crash.rb
[BUG] Segmentation fault at 0x0000000000000038
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux]

Segmentation fault (core dumped)
# gdb `which ruby` core
GNU gdb (Debian 8.2.1-2+b3) 8.2.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/local/bin/ruby...done.
[New LWP 723]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/local/bin/ruby stackprof-crash.rb'.
[...]

PR #200, opened 2023-03-16
Ensure VM is running in signal handler

ianks:
As of a few weeks ago, there's been an increase of sigabrt in prod for us. After hunting the issue down a bit with @peterzhu2118, we honed in on a case where stackprof is signaled and `ruby_current_vm_ptr` is null. This seems to be happening after SIGQUIT'ing a unicorn worker during `rb_during_gc`.

```
* thread #1, name = 'ruby', stop reason = signal SIGABRT
  * frame #0: 0x00007f10787f100b libc.so.6`raise + 203
    frame #1: 0x00007f10787d0859 libc.so.6`abort + 299
    frame #2: 0x0000562e2c372d4f ruby`rb_bug_for_fatal_signal.cold at error.c:776:5
    frame #3: 0x0000562e2c4d56cd ruby`sigsegv(sig=<unavailable>, info=<unavailable>, ctx=<unavailable>) at signal.c:964:5
    frame #4: 0x00007f10789cf420 libpthread.so.0`__restore_rt
    frame #5: 0x0000562e2c396b0b ruby`rb_during_gc at gc.c:11006:12
    frame #6: 0x00007f1070f45391 stackprof.so`stackprof_signal_handler at stackprof.c:745:34
    frame #7: 0x00007f1070f45382 stackprof.so`stackprof_signal_handler(sig=<unavailable>, sinfo=<unavailable>, ucontext=<unavailable>) at stackprof.c:720
    frame #8: 0x00007f10789cf420 libpthread.so.0`__restore_rt
    frame #9: 0x00007f1066be70f4 libselinux.so.1`___lldb_unnamed_symbol8$$libselinux.so.1 + 36
    frame #10: 0x00007f10790daf6b ld-linux-x86-64.so.2`___lldb_unnamed_symbol59$$ld-linux-x86-64.so.2 + 523
    frame #11: 0x00007f10787f48a7 libc.so.6`___lldb_unnamed_symbol191$$libc.so.6 + 247
    frame #12: 0x00007f10787f4a60 libc.so.6`exit + 32
    frame #13: 0x00007f10787d208a libc.so.6`__libc_start_main + 250
    frame #14: 0x0000562e2c373fde ruby`_start + 46
```

Since signal safety is hard to reason about, let's add a sanity check and not assume the Ruby VM is active just because the handler is called.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -12,6 +12,7 @@
 #include <ruby/st.h>
 #include <ruby/io.h>
 #include <ruby/intern.h>
+#include <ruby/vm.h>
 #include <signal.h>
 #include <sys/time.h>
 #include <time.h>
@@ -32,6 +33,7 @@ static const char *fake_frame_cstrs[] = {
 };
 
 static int stackprof_use_postponed_job = 1;
+static int ruby_vm_running = 0;
 
 #define TOTAL_FAKE_FRAMES (sizeof(fake_frame_cstrs) / sizeof(char *))
 
@@ -725,6 +727,11 @@ stackprof_signal_handler(int sig, siginfo_t *sinfo, void *ucontext)
 
     if (!_stackprof.running) return;
 
+    // There's a possibility that the signal handler is invoked *after* the Ruby
+    // VM has been shut down (e.g. after ruby_cleanup(0)). In this case, things
+    // that rely on global VM state (e.g. rb_during_gc) will segfault.
+    if (!ruby_vm_running) return;
+
     if (_stackprof.mode == sym_wall) {
         // In "wall" mode, the SIGALRM signal will arrive at an arbitrary thread.
         // In order to provide more useful results, especially under threaded web
@@ -845,6 +852,12 @@ stackprof_use_postponed_job_l(VALUE self)
     return Qnil;
 }
 
+static void
+stackprof_at_exit(ruby_vm_t* vm)
+{
+    ruby_vm_running = 0;
+}
+
 void
 Init_stackprof(void)
 {
@@ -854,6 +867,9 @@ Init_stackprof(void)
     * See https://github.com/ruby/ruby/commit/0e276dc458f94d9d79a0f7c7669bde84abe80f21
     */
     stackprof_use_postponed_job = RUBY_API_VERSION_MAJOR < 3;
+
+    ruby_vm_running = 1;
+    ruby_vm_at_exit(stackprof_at_exit);
 
 #define S(name) sym_##name = ID2SYM(rb_intern(#name));
     S(object);


ivoanjo:
Perhaps adding an `at_exit { StackProf.stop }` would be a possible solution? I've ran into this in #157 as well.

peterzhu2118:
On Ruby 2.4 and before, it should be called `ruby_current_vm` rather than `ruby_current_vm_ptr`. But at any rate, we shouldn't rely on `ruby_current_vm_ptr` being available for extensions because it isn't marked as public (it isn't wrapped within a `RUBY_SYMBOL_EXPORT_BEGIN`/`RUBY_SYMBOL_EXPORT_END` region). AFAIK there isn't a public API to determine if the Ruby VM is available.

ianks:
> Perhaps adding an `at_exit { StackProf.stop }` would be a possible solution? I've ran into this in #157 as well.

That may help, and is probably a good change in its own right. I would still like to add this guard though since it improves the async signal safety of the handler so it's tricky to reason about.

tenderlove:
> Perhaps adding an `at_exit { StackProf.stop }` would be a possible solution? I've ran into this in #157 as well.

I think stackprof should probably do this by default.  Either that or its callees should check for existence of the VM.

We could check for the VM in stackprof, but that seems kind of fraught as the global is private which means it can be renamed at any time and the compiler is free to do whatever it wants with the symbol name.

ianks:
> We could check for the VM in stackprof, but that seems kind of fraught as the global is private which means it can be renamed at any time and the compiler is free to do whatever it wants with the symbol name.

Yeah I have similar feelings, and very open to other ways of solving this. Maybe a null check in `rb_during_gc` would be better?

peterzhu2118:
> Maybe a null check in rb_during_gc would be better?

No I don't think that will solve the underlying problem since the code in stackprof that is crashing will then instead crash a few lines down in places like `rb_postponed_job_register_one` if the VM doesn't exist.

https://github.com/tmm1/stackprof/blob/52d1df6502eaede5061a2b5d28026e982bcf262a/ext/stackprof/stackprof.c#L745-L764

ianks:
> > Maybe a null check in rb_during_gc would be better?
> 
> No I don't think that will solve the underlying problem since the code in stackprof that is crashing will then instead crash a few lines down in places like `rb_postponed_job_register_one` if the VM doesn't exist.
> 
> https://github.com/tmm1/stackprof/blob/52d1df6502eaede5061a2b5d28026e982bcf262a/ext/stackprof/stackprof.c#L745-L764

I think you are right.

A good stop gap may be to register a new `at_exit` proc with our own variable (`bool ruby_vm_exited`) that we can test against, rather than using the VM*

ianks:
FWIW, I think the failing test may be flaky? I noticed it failing on `master` as well locally (on 3.2).

PR #213, opened 2023-07-08
Add raw line numbers for raw mode

tenderlove:
This commit records line numbers in raw mode along with the frame information.

Before this commit, stackprof would lose information about callee frames at a particular line.  For example, you could not answer "given a frame and line, what function do we call in to at that line?"  This commit encodes the line information along with the raw frame information so that we can answer that question.

This is probably TMI, but when we ask the Ruby frame profiler, it returns a list of memory addresses (CMEs or possibly iseqs?).  AArch64 systems guarantee the top 16 bits won't be used, x86 doesn't use them either (but possibly could in the future, but probably not). Armed with this information, we just put the line number in those top 16 bits and we don't need to allocate any extra memory when doing a profile.

For backwards compatibility, this patch splits apart the information in the `.result` method so existing tools should just work.

### DIFF
--- a/ext/stackprof/stackprof.c
+++ b/ext/stackprof/stackprof.c
@@ -102,7 +102,7 @@ static struct {
     VALUE metadata;
     int ignore_gc;
 
-    VALUE *raw_samples;
+    uint64_t *raw_samples;
     size_t raw_samples_len;
     size_t raw_samples_capa;
     size_t raw_sample_index;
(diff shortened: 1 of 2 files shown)


tenderlove:
With this patch I was able to translate stackprof profiles in to FireFox profiler compatible profiles:

profile


Translation script is [here](https://gist.github.com/tenderlove/75d005ed375556c504716e39dbae0ef9), and I uploaded the sample profile [here](https://profiler.firefox.com/public/ncazp4cwkmk125s3vkdhhpax0r0wrv2g493nf08/calltree/?globalTrackOrder=0&thread=0&timelineType=stack&v=9)

